PR #24, opened 2015-09-18
Add 'send break functionality' with -b command line switch

PrzemekWirkus:
# Description
As a developer
I want to reset mbed-enabled device from command line and see serial port output
on my console
So that I can quickly verify if my device is flashed with proper software

Note: Serial port dump will stop after "{end}' terminator is found

# Example
Users can:
* Send standard serial port BREAK (-b switch) and reset device:
```
$ mbedhtrun -p COM144 -b
```
* Specify other reset plugin:
```
$ mbedhtrun -p COM144 -b -r stlink
```
* Specify baudrate and timeout for serial port:
```
$ mbedhtrun -p COM144:115200 -b
$ mbedhtrun -p COM144:115200:1 -b
```
* Use verbose mode to get more output (-v switch)
```
$ mbedhtrun -p COM144 -b -v
```

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -26,6 +26,7 @@ Write your own programs (import this package) or use 'mbedhtrun' command line to
 """
 
 import sys
+import json
 from optparse import OptionParser
 
 import host_tests_plugins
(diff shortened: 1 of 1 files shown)


jupe:
is there configuration for data-bit and stop-bit counts? what about flow control?

PrzemekWirkus:
Timeout is in seconds.

PrzemekWirkus:
NO data bits or stop bits in this but we can add it if we need it! :)

Issue #48, opened 2016-01-07
Rename rampUp/Down to setup/teardown


niklas-arm:
See https://github.com/ARMmbed/htrun/issues/41#issuecomment-165783839
and also https://github.com/ARMmbed/mbed-clitest/pull/207.

0xc0170:
+1 to have unified naming for these functions

PrzemekWirkus:
@stegut01 Please implement this simple issue! :)

ciarmcom:
ARM Internal Ref: IOTSYST-740

PrzemekWirkus:
We've merged this change already. We will release soon so you will be able to change your pending PRs using setup/teardown.

BTW, we still didn't make setup->test->teardown robust so it might not work as expected.

PrzemekWirkus:
I will close this because issue is about renaming not actually making teardown() work.

PR #50, opened 2016-01-08
[issue 48] Rename rampUp/Down to setup/teardown

stegut01:
Fixes issue #48

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -393,16 +393,16 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         # Read serial and wait for binary execution end
         try:
             self.test_supervisor = get_host_test("run_binary_auto")
-            # Call to rampUp if function is implemented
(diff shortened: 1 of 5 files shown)


0xc0170:
One question , I noticed all functions in this module are using lower case names, why rumpUp and Down are an exception? now setUp/tearDown. I assume it's to be aligned with other tools (I noticed in cli test they use camel case for functions). Which makes a question of if this compatibility makes sense (adding new API will follow which one?), either to use one or another?

PrzemekWirkus:
Like I said, rampUp/rampDown was "copied" from Jussi's framework and names there were camel case.
We will align with module's (Pythonic) coding styles.

PrzemekWirkus:
@stegut01 Stefan, please rename ```tearDown``` to ```teardown```, see [coding styles](https://github.com/ARMmbed/greentea/blob/master/docs/CONTRIBUTING.md#naming) for our tools.

PrzemekWirkus:
+1

0xc0170:
:clap:

PR #54, opened 2016-01-12
Aborting mechanism when parent process wants to stop mbedhtrun

mazimkhan:
Fix for issue https://github.com/ARMmbed/greentea/issues/48

Changes:
 - Added a stdin monitoring thread
 - Aborting selector when stdin is closed.
 - Check for abort at different steps of execution.
 - Selector aborts test case.
 - Added new based class method abort().

Creating PR for feedback from issue creator. Unit tests will be added soon.

This PR is complimentary to https://github.com/ARMmbed/greentea/pull/63 (they should come at the same time in release.

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -512,12 +512,9 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         self.notify("HOST: Aborted by parent process!")
         self.aborted = True
         try:
-            if self.test_supervisor is not None and 'abort' in dir(self.test_supervisor) and \
-                    callable(getattr(self.test_supervisor, 'abort')):
-                self.test_supervisor.abort()
+            self.test_supervisor.rampDown()
         finally:
-            pass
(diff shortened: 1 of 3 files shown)


autopulated:
From some experience trying to do things like this cleanly: the handling of file polling can vary substantially between platforms, so it's worth double-checking that this works as intended across windows, mac and linux!

mazimkhan:
@autopulated thanks for reminding. I have cross platform testing in mind. This PR is more for @0xc0170 to test in his environment. 
I aware that read may break for EAGAIN and EINTR on Linux. Havn't used mac ever. Any specific things I should be careful about?

PrzemekWirkus:
@mazimkhan BTW, we've replaced rampUp/rampDown with setup/teardown.

mazimkhan:
@PrzemekWirkus @stegut01   please review this PR as well.

PrzemekWirkus:
I will review it today! :)

Issue #56, opened 2016-01-13
pySerial 3.0.1 version breaks unit tests


PrzemekWirkus:
We falsely assumed pySerial will have 3.x version numbers.
* Time to update pySerial detection function and unit test to guard it.
```
======================================================================
FAIL: test_pyserial_version_as_float (test.host_test_plugins.HostOSDetectionTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\work\pypi\htrun\test\host_test_plugins.py", line 61, in test_pyserial_version_as_float
    self.assertTrue(self.re_float.findall(pyserial_version))
AssertionError: [] is not true

----------------------------------------------------------------------
Ran 22 tests in 0.034s
```

ciarmcom:
ARM Internal Ref: IOTSYST-759

PR #57, opened 2016-01-13
Add better detection of pySerial 3.x.y libraries

PrzemekWirkus:
Fix for https://github.com/ARMmbed/htrun/issues/56.

We falsely assumed pySerial will have 3.x version numbers. So it was parsed to float, now with 3.0.1 version we got into trouble.
This PR tries to fix that...

### DIFF
--- a/mbed_host_tests/host_tests/default_auto.py
+++ b/mbed_host_tests/host_tests/default_auto.py
@@ -35,8 +35,8 @@ class DefaultAuto(BaseHostTest):
 
                 line += c
                 if '\n' in line:
-                    if '{end}' in line:
-                        return None # Run executable end
+                    #if '{end}' in line:
+                    #    return None # Run executable end
                     i = line.index('\n')
                     line = line[i:]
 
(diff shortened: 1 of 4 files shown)


0xc0170:
Why are we keeping both 2x and 3.x in this module?

What's the benefit of having this code in, besides that it's compatible with both. Is it worth staying with 2.x for our future releases?
```
         if self.is_pyserial_v3:
-            return self.__safe_sendBreak_v3_0(serial)
+            return self._safe_sendBreak_v3_0(serial)
-        return self.__safe_sendBreak_v2_7(serial)
+        return self._safe_sendBreak_v2_7(serial)
```

PrzemekWirkus:
@0xc0170 It is for backward compatibility, because we have some CI machines with pySerial 2.7 installed and we do not want also used who are dependant on 2.7 or 3.0 to choose from one or migrate to virtual env.

So this will be removed in the future, is added now as temporary thin. Because pySerial 3.0 is a new implementation. We want to give people time to migrate when they want.

PR #63, opened 2016-01-20
Add missing setup/teardown call

### LINKED ISSUE #60: Missing setup() / teardown() call in main htrun flow
# Description
We are [only calling ```setup()``` / ```teardown()```](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/__init__.py#L413) when we handle ```---run``` switch (not used in CI). We are not calling setup/teardown in normal workflow:
```python
def execute_run(self):
...
try:
    self.test_supervisor = get_host_test("run_binary_auto")
    # Call to setup if function is implemented
    if hasattr(self.test_supervisor, 'setup') and callable(getattr(self.test_supervisor, 'setup')):
        self.test_supervisor.setup()

    # Call to test function
    result = self.test_supervisor.test(self)    # This is blocking, waits for {end}

    # Call to teardown if function is implemented
    if hasattr(self.test_supervisor, 'teardown') and callable(getattr(self.test_supervisor, 'teardown')):
        self.test_supervisor.teardown()
except Exception, e:
    print str(e)
    self.print_result(self.RESULT_ERROR)
```

We should be calling it also [here](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/__init__.py#L480):
```python
def execute(self):
...
try:
    CONFIG = self.detect_test_config(verbose=True) # print CONFIG

    result = None
    if "host_test_name" in CONFIG:
        if is_host_test(CONFIG["host_test_name"]):
            #self.notify("HOST: CONFIG['host_test_name'] is '%s'" % CONFIG["host_test_name"])
            self.test_supervisor = get_host_test(CONFIG["host_test_name"])
            result = self.test_supervisor.test(self)    #result = self.test()
        else:
            self.notify("HOST: Error! Unknown host test name '%s' (use 'mbedhtrun --list' to verify)!"% CONFIG["host_test_name"])
            self.notify("HOST: Error! You can use switch '-e <dir>' to specify local directory with host tests to load")
            self.print_result(self.RESULT_ERROR)
    else:
        self.notify("HOST: Error! No host test name defined in preamble")
        self.print_result(self.RESULT_ERROR)
```

probably missing code in above snippet:
```python
self.test_supervisor.setup()
result = self.test_supervisor.test(self)    #result = self.test()
self.test_supervisor.teardown()
```

# Reproduce
* In [mbed_host_tests/host_tests/default_auto.py](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests/default_auto.py#L21) add setup/teardown with prints:
```python
class DefaultAuto(BaseHostTest):
    """ Simple, basic host test's test runner waiting for serial port
        output from MUT, no supervision over test running in MUT is executed.
    """
    def setup(self):
        stdout.write("!!!")
        stdout.flush()
        
    def teardown(self):
        stdout.write("@@@")
        stdout.flush()
    
    def test(self, selftest):
        stdout.write("%%%")
        stdout.flush()
        return "XXX"
```

* Add calls to setup/teardown [here](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/__init__.py#L488):
```python
                    self.test_supervisor = get_host_test(CONFIG["host_test_name"])
                    self.test_supervisor.setup()
                    result = self.test_supervisor.test(self)    #result = self.test()
                    self.test_supervisor.teardown()
```

* Call ```mbedgt```:
```
$ mbedgt -V -n mbed-drivers-test-basic
...
mbedgt: selecting test case observer...
        calling mbedhtrun: mbedhtrun -d F: -p COM184:9600 -f ".\build\frdm-k64f-gcc\test\mbed-drivers-test-basic.bin" -C 4 -c shell -m K64F -e "./test/host_tests"
mbedgt: mbed-host-test-runner: started
HOST: Instrumentation: "COM184" and disk: "F:"
HOST: My PID is 7516
HOST: Copy image onto target...
        1 file(s) copied.
HOST: Initialize serial port...
...port ready!
HOST: Reset target...
HOST: Detecting test case properties...
HOST: Property 'timeout' = '20'
HOST: Property 'host_test_name' = 'default_auto'
HOST: Property 'description' = 'Basic'
HOST: Property 'test_id' = 'MBED_A1'
HOST: Start test...
!!!%%%@@@
{{XXX}}
{{end}}
```

stegut01:
fix #60

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -222,7 +222,6 @@ class DefaultTestSelector(DefaultTestSelectorBase):
     def __init__(self, options):
         """! ctor
         """
-        self.result = None
         self.options = options
 
         # Handle extra command from
@@ -307,8 +306,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
 
         serial_port.flush()
         # Reset using one of the plugins
-        self.result = host_tests_plugins.call_plugin('ResetMethod', reset_type, serial=serial_port, disk=disk)
-        if not self.result:
+        result = host_tests_plugins.call_plugin('ResetMethod', reset_type, serial=serial_port, disk=disk)
+        if not result:
             print "mbedhtrun: reset plugin failed"
             print json.dumps({
                 "port" : port,
@@ -382,8 +381,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         # Copy image to device
         if self.options.skip_flashing is False:
             self.notify("HOST: Copy image onto target...")
-            self.result = self.mbed.copy_image()
-            if not self.result:
+            result = self.mbed.copy_image()
+            if not result:
                 self.print_result(self.RESULT_IOERR_COPY)
         else:
             self.notify("HOST: Image copy onto target SKIPPED!")
@@ -392,8 +391,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
             return
         # Initialize and open target's serial port (console)
         self.notify("HOST: Initialize serial port...")
-        self.result = self.mbed.init_serial()
-        if not self.result:
+        result = self.mbed.init_serial()
+        if not result:
             self.print_result(self.RESULT_IO_SERIAL)
 
         if self.aborted:
@@ -401,8 +400,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         # Reset device
         if self.options.skip_reset is False:
             self.notify("HOST: Reset target...")
-            self.result = self.mbed.reset()
-            if not self.result:
+            result = self.mbed.reset()
+            if not result:
                 self.print_result(self.RESULT_IO_SERIAL)
         else:
             self.notify("HOST: Target reset SKIPPED!")
@@ -433,8 +432,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
             return
         if self.options.skip_flashing is False:
             self.notify("HOST: Copy image onto target...")
-            self.result = self.mbed.copy_image()
-            if not self.result:
+            result = self.mbed.copy_image()
+            if not result:
                 self.print_result(self.RESULT_IOERR_COPY)
                 return  # No need to continue, we can't flash device
         else:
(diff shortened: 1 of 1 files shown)


PrzemekWirkus:
+1

mbed_host_tests/__init__.py
```diff
@@ -501,6 +495,8 @@ def execute(self):
         except Exception, e:
             print str(e)
             self.print_result(self.RESULT_ERROR)
+        finally:
+            self.test_supervisor.teardown()
```
mazimkhan: It can raise self.test_supervisor used before assignment. I think there should be a try except finally block around 

```python
self.test_supervisor.setup()
result = self.test_supervisor.test(self)
```

PrzemekWirkus: Yes, ```get_host_test()``` can return ```None``` so we should either check for:
```python
if self.test_supervisor:  # self.test_supervisor can be None
    # ...
```
or do try/catch around it.

mbed_host_tests/__init__.py
```diff
@@ -494,14 +479,25 @@ def execute(self):
...
                 self.notify("HOST: Passive mode...")
         except Exception, e:
             print str(e)
             self.print_result(self.RESULT_ERROR)
 
+    def run_test(self, host_test_name):
+        try:
+            self.test_supervisor = get_host_test(host_test_name)
+            self.test_supervisor.setup()
+            self.result = self.test_supervisor.test(self)    # This is blocking, waits for {end}
```
mazimkhan: This function can return the result in the end. We should put less variables in the object state as they would need maintenance.

mbed_host_tests/__init__.py
```diff
@@ -494,14 +479,25 @@ def execute(self):
...
+                self.print_result(self.result)
             else:
                 self.notify("HOST: Passive mode...")
         except Exception, e:
             print str(e)
             self.print_result(self.RESULT_ERROR)
 
+    def run_test(self, host_test_name):
+        try:
+            self.test_supervisor = get_host_test(host_test_name)
```
mazimkhan: Since ```if is_host_test(CONFIG["host_test_name"]):``` checks that host test exists before ```run_test ()``` calls ```get_host_test()``` it should always work. But at this line if feels unsafe to call ```test_supervisor.setup()``` without checking.
Perhaps we can call ```get_host_test()``` only and before calling ```run_test(test_supervisor)``` or put an assert here like
``` assert self.test_supervisor is not None, "get_host_test() return None"```

mazimkhan:
Code looks good
+1

PR #65, opened 2016-02-02
[devel_transport] Async features

PrzemekWirkus:
# Description 
This changes introduce async ```htrun``` features:
* KiVi - key-value protocol, see: ```{{key;value}}}``` protocol.
* ```sync``` message to sync/handshake DUT and host test.
* Async serial port event queue with callback mappings.
* Callbacks registration.
* Ported few host tests to new async:
  * Example: default_auto, echo, hello_auto, rtc_auto and (I think more).

# Limitations
This feature is not compatible with Greentea in current state, please use ```mbedhtrun``` command separately.

# Example usage:
```
$ git clone <this htrun-branch>
$ cd htrun
$ python setup.py develop
```

You can build yourself ```mbed-drivers``` from corresponding PR.
```
$ git clone <mbed-drivers devel_greentea>
$ yt target
$ yt build
$ mbedgt --list
```
Run some test with ```mbedhtrun``` (not yet with ```mbedgt```!!!)
```
$ mbedhtrun -d F: -p COM184:9600 -f ".\build\frdm-k64f-gcc\test\mbed-drivers-test-echo.bin" -C 4 -c shell -m K64F
```
Where ```-d F:``` specifies mount point of your e.g. ```K64F```, serial port ```-p COM184:9600```.

You should see:
```
$ mbedhtrun ...
[1454445483.33][INF] copy image onto target... SKIPPED!
[1454445483.34][INF] starting host test process...
[1454445483.94][INF] starting connection process...
[1454445483.94][INF] Serial port=COM184 baudrate=9600 program_cycle_s=4
[1454445483.94][INF] reset device (send_break(0.50 sec))
[1454445484.44][INF] wait after flashing (sleep(4.00 sec))
[1454445488.45][INF] sending preamble 'd238e9ec-65d4-4b60-a141-a1e146578319'...
[1454445488.65][RXD] {{sync;d238e9ec-65d4-4b60-a141-a1e146578319}}
[1454445488.65][RXD] {{timeout;15}}
[1454445488.65][RXD] {{host_test_name;echo}}
[1454445488.65][INF] found KV pair in stream: {{sync;d238e9ec-65d4-4b60-a141-a1e146578319}}, queued...
[1454445488.65][INF] sync KV found, uuid=d238e9ec-65d4-4b60-a141-a1e146578319, timestamp=1454445488.646000
[1454445488.65][INF] found KV pair in stream: {{timeout;15}}
[1454445488.65][INF] setting timeout to: 15 sec
[1454445488.65][INF] found KV pair in stream: {{host_test_name;echo}}
[1454445488.65][INF] host test setup() call...
[1454445488.65][TXD] {{echo;c9e15677-7713-480b-9dfa-4df7c6ab5fde}}
[1454445488.65][INF] CALLBACKs updated
[1454445488.65][INF] host test detected: echo
[1454445488.85][RXD] {{echo;c9e15677-7713-480b-9dfa-4df7c6ab5fde}}
```


mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -15,28 +15,116 @@
...
+        if self.event_queue:
+            self.dut_event_queue.put((key, value, time()))
+
+    def send_kv(self, key, value):
+        self._notify_dut(key, value)
+
+    def setup(self):
+        pass
+
+    def test(self):
```
0xc0170: What's the function of test() ? Is it required to be implemented or optional? If required, should it raise NotImplemented error here?

mbed_host_tests/host_tests/default_auto.py
```diff
@@ -15,32 +15,32 @@
...
+        self.notify_complete()
+
+    def setup(self):
+        self.register_callback('end', self._callback_end)
+        self.register_callback('exit', self._callback_exit)
+
+    def test(self):
+        return self.result
+
+    def teardown(self):
```
0xc0170: I would say default auto is with only test() implemented, setup/teardown is not default, is it?

Or is it still as you described yesterday, test() is the method called after HostTest receives ```notify_complete()``` ?

I would imagine test() to be

```
def test(self):
   # setting timeout for this test case , it's not the same as host test timeout set by a client which is setting timeout for entire test unit. Or?
   self.set_timeout(5);
   self.register_callback(.....)
   # here I return, a timeout is set, means this test() is asynch, wait for notify_complete() or timeout. 
   # notify_complete can have an argument which would indicate if it is Pass or Fail
```

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -15,28 +15,116 @@
...
-            teardown()
+    def _assign_callbacks(self):
+        """! Assigns default callback handlers
         """
+        for key in self.printable:
+            self.callbacks[key] = self._callback_forward
+
+    def register_callback(self, key, callback):
+        if type(key) is not str:
+            raise TypeError
```
0xc0170: Are we catching exceptions somewhere? Why not using assert? or TypeError with a message what went wrong?

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -15,28 +15,116 @@
...
+        pass
+
+    def test(self):
+        pass
+
+    def teardown(self):
+        pass
+
+
+class BaseHostTest(HostTestCallbackBase):
```
0xc0170: which of these 3 classes in this file are intended to be used by a user? all 3? why setup,test,teardown are empty in all these classes?

mbed_host_tests/host_tests/dev_null_auto.py
```diff
@@ -19,34 +19,23 @@
...
+    def _callback_result(self, key, value, timestamp):
+        # We should not see result data in this test
+        self.result = False
+
+    def _callback_to_stdout(self, key, value, timestamp):
+        self.result = True
+
+    def setup(self):
+        # This should never appear in this test
+        self.register_callback("end", self._callback_result)
```
0xc0170: > This should never appear in this test

means end  , exit and to_null are reserved keywords, which I assume they are reserved? However I can't find any reference to these keys in this PR. register_callback does not seem to object for these  keys ?

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -0,0 +1,143 @@
...
+        self.config = config
+        self.logger = HtrunLogger(prn_lock)
+        program_cycle_s = int(self.config['program_cycle_s']) if 'program_cycle_s' in self.config else 4
+
+        self.logger.prn_inf("Serial port=%s baudrate=%s program_cycle_s=%s"% (str(port), (baudrate), (program_cycle_s)))
+        try:
+            self.serial = Serial(port, baudrate=baudrate, timeout=0.2)
+        except SerialException as e:
+            self.logger.prn_err(str(e))
+            self.serial = None
```
0xc0170: if Serial() fails , does it make sense go forward and do send_break and program_cycle_wait() ? send_break does not make sense if serial is None. program cycle might stay.

mbed_host_tests/__init__.py
```diff
@@ -430,91 +357,126 @@ def execute(self):
...
+        p.start()
+
+        consume_preamble_events = True
+        while (time() - start_time) < timeout_duration:
+            # Handle default events like timeout, host_test_name, ...
+            if event_queue.qsize():
+                (key, value, timestamp) = event_queue.get()
+
+                if consume_preamble_events:
+                    if key == 'timeout':
```
0xc0170: Where are these keys described? I see timeout, host_test_name, sync. Are they coming from the client side? should they be defined here as well in this code base? If I find this print "sync KV found" , I have no clue what it means from reading this code

test/host_test_scheme.py
```diff
@@ -59,18 +59,12 @@ def test(self, selftest)
...
-        for i, ht_name in enumerate(self.HOSTREGISTRY.HOST_TESTS):
-            ht = self.HOSTREGISTRY.HOST_TESTS[ht_name]
-            if ht is not None and hasattr(ht, 'test') and callable(getattr(ht, 'test')) and ht.test.func_code.co_argcount == 2:
-                self.assertEqual(('self', 'selftest'), ht.test.func_code.co_varnames[:ht.test.func_code.co_argcount])
+            if ht is not None and hasattr(ht, 'setup') and callable(getattr(ht, 'setup')):
+                self.assertEqual(1, ht.setup.func_code.co_argcount)
+            #if ht is not None and hasattr(ht, 'test') and callable(getattr(ht, 'test')):
+            #    self.assertEqual(1, ht.test.func_code.co_argcount)
+            #if ht is not None and hasattr(ht, 'teardown') and callable(getattr(ht, 'teardown')):
+            #    self.assertEqual(1, ht.teardown.func_code.co_argcount)
```
0xc0170: why is this commented out?

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python
+"""
+mbed SDK
+Copyright (c) 2011-2015 ARM Limited
```
0xc0170: year should be 2016 in some files in this PR, and we might consider updating the text as this one comes from mbed SDK (line 3)

mbed_host_tests/__init__.py
```diff
@@ -430,91 +357,126 @@ def execute(self):
...
+                        self.logger.prn_inf("setting timeout to: %d sec"% int(value))
+                        start_time = time()
+                        timeout_duration = int(value) # New timeout
+                    elif key == 'host_test_name':
+                        self.test_supervisor = get_host_test(value)
```
0xc0170: ``` test_supervisor ``` is actually current_host_test ?

mbed_host_tests/__init__.py
```diff
@@ -430,91 +357,126 @@ def execute(self):
...
+                        if self.test_supervisor:
+                            self.test_supervisor.event_queue = event_queue
+                            self.test_supervisor.dut_event_queue = dut_event_queue
+                            self.test_supervisor.setup()
```
0xc0170: We expect setup() sets  the callbacks, as the condition below checks if there are any callbacks. What does happen if I do:

```
# my host test
def test():
  self.register_callback();
```
Does this fail anyhow? My host test will print an error? registering callback in the test function?

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -15,28 +15,171 @@
...
+    def setup_communication(self, event_queue, dut_event_queue):
+        """! Setup queues used for IPC """
+        self.__event_queue = event_queue         # To main even loop
+        self.__dut_event_queue = dut_event_queue # To DUT
+
+    def setup(self):
+        """! Setup your tests and callbacks """
+        raise NotImplementedError
+
+    def result(self):
```
mazimkhan: Why do we have ```result()``` method when an async test can notify completion by calling ```notify_complete```. Two points to convey test results can cause buggy test implementation.

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -15,28 +15,171 @@
...
+
+        self.__callbacks[key] = callback
+
+    def get_callbacks(self):
+        return self.__callbacks
+
+    def setup(self):
+        pass
+
+    def result(self):
```
mazimkhan: ```result``` returns test results. So it can't return None. Also since test result is not known as this point it should have an implementation in the derived host test. Hence it ```raise NotImplementedError```. I propose to remove it completely. see above comment.

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -15,28 +15,171 @@
...
+        pass
+
+
+class BaseHostTest(HostTestCallbackBase):
+
+    def __init__(self):
+        HostTestCallbackBase.__init__(self)
+
+    def setup(self):
+        pass
```
mazimkhan: If base class implementation is not overridden then ```setup, result and teardown``` can be removed.

mbed_host_tests/host_tests/base_host_test.py
0xc0170: +1 , I am still not about convinced result() neither

PR #66, opened 2016-02-21
Event handler decorator for automatic registering of event callbacks

mazimkhan:
This is an alternative approach to implementing ```setup``` method and calling ```self.register_callback``` for registering event callbacks. 
Now event callbacks can be decorated with a decorator as ```@event_cb('<key>')```. A base implementation of ```setup``` is added in the ```BaseHostTest``` to look for decorated methods and register them as callbacks. Thus host test developers do not need to implement ```setup``` anymore. 

```
Example:
 #        Define a event callback method with @event_callback decorator like:
 
          @event_callback('<event key>')
          def event_handler(self, key, value, timestamp):
             do something..
 
```

### DIFF
--- a/mbed_host_tests/host_tests/base_host_test.py
+++ b/mbed_host_tests/host_tests/base_host_test.py
@@ -81,7 +81,7 @@ class BaseHostTestAbstract(object):
         raise NotImplementedError
 
 
-def event_cb(key):
+def event_callback(key):
     """
     Decorator for defining a event callback method. Adds a property attribute "event_key" with value as the passed key.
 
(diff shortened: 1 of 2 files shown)


mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -80,6 +81,19 @@ def teardown(self):
         raise NotImplementedError
 
 
+def event_cb(key):
```
PrzemekWirkus: Please change from ```event_cb``` to ```event_callback```.
Niklas told me before they prefer this type of names (explicit, so no one had to quess what ```_cb``` stands for).

PrzemekWirkus:
Please add to your PR description example of how to use it.

mbed_host_tests/host_tests/base_host_test.py
0xc0170: +1 for being explicit !

PrzemekWirkus:
Very nice! :)

PR #67, opened 2016-02-22
Default handling for 'end' event

mazimkhan:
For most cases event 'end' callback has following handling:

```python
def end_event_callback(self, key, value, timestamp):
    self.notify_complete(value == 'success')
```

Hence this PR implements a default callback with above handling in base ```class HostTestCallbackBase```.
Users can still over write this handler by creating a handler and registering for key 'end'. Example:

```python
def my_end_callback(self, key, value, timestamp):
    # my implementation for 'end' handling
    ....
```

### DIFF
--- a/mbed_host_tests/host_tests/base_host_test.py
+++ b/mbed_host_tests/host_tests/base_host_test.py
@@ -117,12 +117,28 @@ class HostTestCallbackBase(BaseHostTestAbstract):
         ]
 
         self.__assign_default_callbacks()
+        # Register default handler for event 'end' before assigning user defined callbacks to let users over write it.
+        self.register_callback('end', self.default_end_callback)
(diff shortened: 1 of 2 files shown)


mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -117,6 +117,8 @@ def __init__(self):
         ]
 
         self.__assign_default_callbacks()
+        # Register default handler for event 'end' before assigning user defined callbacks to let users over write it.
+        self.register_callback('end', self.__default_end_callback)
```
PrzemekWirkus: This belongs to body of 
```python
def __assign_default_callbacks():
   ...
```

PrzemekWirkus:
Very cool! :)

PR #75, opened 2016-03-15
Changes to support DUT reset in a test

mazimkhan:
- Support to reset DUT in test.
- Now greentea will provide mbed platform's unique target id to confirm same device is used after reset.

Please see following sequence diagram that illustrates how a test can manage its state before a after restart:

```
DUT                     host
---                     ----
 |    {{__sync;UUID}}     |
 |<-----------------------|
 |                        |
 |    {{__sync;UUID}}     |
 |----------------------->|

 - - - - - - - - - - -  - -
        HANDSHAKE !
 - - - - - - - - - - -  - -

 |                        |
 |   {{__timeout;%s}}     |
 |----------------------->|
 |                        |
 |{{__host_test_name;%s}} |
 |----------------------->|

 - - - - - - - - - - -  - -
 TEST SUITE CONTROL UNDER 
         HOST TEST 
 - - - - - - - - - - -  - -     
                                  ht = HostTest(__host_test_name)
 |                        |       ht.setup()

 - - - - - - - - - - -  - - - - - -
          TEST CASE FLOW
 - - - - - - - - - - -  - - - - - -
 - - - - - - - - - - -  - -
 Since Host Test only responds to events. It is essential that target initiates
test by sending first event/key-value pair.
 - - - - - - - - - - -  - -     

 | {{start;value}}        |       |
 | -------------------->  |       |
 |          .             |       |  
 |          .             |       - - - - - - - - - -
 |          .             |        Host test responds to the 'start' event and sends back default target state.
 |          .             |       - - - - - - - - - - 
 | {{state;x}}            |       |  
 | <--------------------  |       |  
 |                        |       |  
 | {{echo;1}}             |       |
 | -------------------->  |       |
 | {{echo;1}}             |       |  
 | <--------------------  |       |  
 |          .             |       - - - - - - - - - -
 |          .             |        Target sends n echo events and expect n echoes back
 |          .             |       - - - - - - - - - - 
 | {{state;y}}            |       |
 | -------------------->  |       |
 | {{reset;0}}            |       |
 | -------------------->  |       |
 |          .             |       - - - - - - - - - -
 |          .             |        Target switches state and informs Host test. Then requests a reset
 |          .             |       - - - - - - - - - - 
 - - - - - - - - - - -  - -
 On event 'reset' host test sends event '__reset' that is internally handled by htrun event loop. 
It performs a software reset. On start-up target again performs sync handshake that is handled 
in the same way as before by htrun. Target then sends "start" event. On which host test sends its
current state.
[...]

PR #77, opened 2016-03-29
Hardware reset feature

mazimkhan:
Extension on PR https://github.com/ARMmbed/htrun/pull/75 for hardware reset. Please see feature documentation in https://github.com/ARMmbed/htrun/pull/75

Changes:

Now ```reset_dut(reset_type)``` accepts an argument ```reset_type``` that can have two string value ```software_reset``` or ```hardware_reset```. 
* ```software_reset``` resets the device by sending break command over serial.
* ```hardware_reset``` uses mbed-tas-rm service to reset the device.

Target test can initiate software or hardware reset by sending ```software_reset``` or ```hardware_reset``` as value with a corresponding event key. Target can either send a host test handled key for which host test calls ```reset_dut```. Or target can directly send ```{{__reset_dut;<reset type>}}``` and htrun event loop will reset the target.

For hardware reset htrun requires that environment variables ```MBED_TAS_RM_IP``` and ```MBED_TAS_RM_PORT``` are set. In Jenkins these environment variables are set in Jenkins->Configure->Global environment settings. On desktop user would need to set these environment variables pointing to locally running mbed-tas-rm service. See https://github.com/ARMmbed/mbed-tas-rm documentation.

### DIFF
--- a/mbed_host_tests/host_tests/base_host_test.py
+++ b/mbed_host_tests/host_tests/base_host_test.py
@@ -40,7 +40,6 @@ class BaseHostTestAbstract(object):
     def __notify_dut(self, key, value):
         """! Send data over serial to DUT """
         if self.__event_queue:
-            print "adding to dut event queue %s %s" % (key, value)
             self.__dut_event_queue.put((key, value, time()))
 
     def notify_complete(self, result=None):
--- a/mbed_host_tests/host_tests/base_host_test.py
+++ b/mbed_host_tests/host_tests/base_host_test.py
@@ -40,6 +40,7 @@ class BaseHostTestAbstract(object):
     def __notify_dut(self, key, value):
         """! Send data over serial to DUT """
         if self.__event_queue:
+            print "adding to dut event queue %s %s" % (key, value)
             self.__dut_event_queue.put((key, value, time()))
 
     def notify_complete(self, result=None):
@@ -49,13 +50,13 @@ class BaseHostTestAbstract(object):
         if self.__event_queue:
             self.__event_queue.put(('__notify_complete', result, time()))
 
-    def reset_dut(self):
+    def reset_dut(self, value):
         """
         Reset device under test
         :return:
         """
         if self.__event_queue:
-            self.__event_queue.put(('__reset_dut', True, time()))
+            self.__event_queue.put(('__reset_dut', value, time()))
 
     def notify_conn_lost(self, text):
         """! Notify main even loop that there was a DUT-host test connection error
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -93,8 +93,7 @@ class SerialConnectorPrimitive(object):
     def write_kv(self, key, value):
         kv_buff = "{{%s;%s}}\n"% (key, value)
         self.write(kv_buff)
-        print "Written: %s" % kv_buff
-        #self.logger.prn_txd(kv_buff)
+        self.logger.prn_txd(kv_buff)
         return kv_buff
 
     def flush(self):
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -35,7 +35,7 @@ class SerialConnectorPrimitive(object):
         self.logger = HtrunLogger(prn_lock, 'SERI')
         self.LAST_ERROR = None
 
-        self.logger.prn_inf("serial(port=%s, baudrate=%d)"% (self.port, self.baudrate))
+        self.logger.prn_inf("serial(port=%s, baudrate=%d, timeout=%s)"% (self.port, self.baudrate, self.timeout))
         try:
             self.serial = Serial(port, baudrate=baudrate, timeout=self.timeout)
         except SerialException as e:
@@ -93,7 +93,8 @@ class SerialConnectorPrimitive(object):
     def write_kv(self, key, value):
         kv_buff = "{{%s;%s}}\n"% (key, value)
         self.write(kv_buff)
-        self.logger.prn_txd(kv_buff)
+        print "Written: %s" % kv_buff
+        #self.logger.prn_txd(kv_buff)
         return kv_buff
 
     def flush(self):
@@ -174,11 +175,11 @@ def conn_process(event_queue, dut_event_queue, prn_lock, config):
             break
 
         # Send data to DUT
-        if not dut_event_queue.empty():
-            try:
-                (key, value, _) = dut_event_queue.get(timeout=1)
-            except QueueEmpty:
-                continue
+        try:
+            (key, value, _) = dut_event_queue.get(timeout=0)
+        except QueueEmpty:
+            pass # Check if target sent something
+        else:
             # Return if state machine in host_test_default has finished to end process
             if key == '__host_test_finished' and value == True:
                 connector.finish()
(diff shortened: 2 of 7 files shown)


PrzemekWirkus:
@mazimkhan I think we could move HW reset funcitonality to e.g. htrun.plugins:
https://github.com/ARMmbed/htrun/tree/master/mbed_host_tests/host_tests_plugins

So we can call it from plugin API? I did not explore this opportunity but it should be done the same way as it was here: https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py#L59

mazimkhan:
@PrzemekWirkus sure i will explore that.

mazimkhan:
@PrzemekWirkus @stegut01 please review. Power cycle code is moved to a new plugin.

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -49,13 +49,13 @@ def notify_complete(self, result=None):
         if self.__event_queue:
             self.__event_queue.put(('__notify_complete', result, time()))
 
-    def reset_dut(self):
+    def reset_dut(self, value):
```
PrzemekWirkus: This breaks reset test in greentea-client ?

mbed_host_tests/host_tests_plugins/module_power_cycle_mbed.py
```diff
@@ -0,0 +1,179 @@
...
+        Performs hardware reset of target ned device.
+
+        :return:
+        """
+        result = False
+        try:
+            ip = os.environ['MBED_TAS_RM_IP']
+            port = os.environ['MBED_TAS_RM_PORT']
+        except KeyError, e:
+            print "HOST: Failed read environment variable (" + str(e) + "). Can't perform hardware reset."
```
PrzemekWirkus:
```python
self.print_plugin_error("HOST: Failed read environment variable (" + str(e) + "). Can't perform hardware reset.")
```

mbed_host_tests/host_tests_plugins/module_power_cycle_mbed.py
```diff
@@ -0,0 +1,179 @@
...
+                }
+            ]
+        }
+
+        result = False
+
+        # reset target
+        switch_off_req = self.__run_request(ip, port, switch_off_req)
+        if switch_off_req is None:
+            print "HOST: Failed to communicate with TAS RM!"
```
PrzemekWirkus:
```python
self.print_plugin_error("HOST: Failed to communicate with TAS RM!")
```

mbed_host_tests/host_tests_plugins/module_power_cycle_mbed.py
```diff
@@ -0,0 +1,179 @@
...
+            return result
+
+        if "error" in switch_off_req['sub_requests'][0]:
+            print "HOST: Failed to reset target. error = %s" % switch_off_req['sub_requests'][0]['error']
```
PrzemekWirkus: please use ```self.print_plugin_error()```

mbed_host_tests/host_tests_plugins/module_power_cycle_mbed.py
```diff
@@ -0,0 +1,179 @@
...
+        if self.check_parameters(capability, *args, **kwargs) is True:
+            if capability in HostTestPluginPowerCycleResetMethod.capabilities:
+                target_id = kwargs['target_id']
+                device_info = kwargs['device_info']
+                result = self.__hw_reset(target_id, device_info)
+        return result
+
+    def __hw_reset(self, target_id, device_info):
+        """
+        Performs hardware reset of target ned device.
```
PrzemekWirkus: Typo?

mbed_host_tests/host_tests_plugins/module_power_cycle_mbed.py
```diff
@@ -0,0 +1,179 @@
...
+
+        :return:
+        """
+        result = False
+        try:
```
PrzemekWirkus: You probably wanna check if these env variables are set before you actually allow people to reset? 
It is maybe not that important but it would be cool if ```__hw_reset()``` is not depending on these env.variables. Rather they are passed as parameters from ```execute()```.
Rationale: It would be easier to use ```__hw_reset()``` as API function so I can call it with IP and port and do not rely on env. settings.

mazimkhan: Since the plugin is mbed-tas-rm specific it does not matter. Outcome will be the same. It is good to encapsulate everything mbed-tas-rm specific under this plugin. I do believe the name is too generic. Perhaps because it is first plugin for power cycling.

mbed_host_tests/host_tests/base_host_test.py
mazimkhan: Yes. It is a breaking change. Please see the corresponding PR for greentea-client test https://github.com/ARMmbed/greentea-client/pull/10.

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -190,10 +192,24 @@ def start_conn_process():
                             result = value
                             break
                         elif key == '__reset_dut':
-                            # Disconnecting and re-connecting comm process will reset DUT
+                            # Disconnect to avoid connection lost event
                             dut_event_queue.put(('__host_test_finished', True, time()))
```
stegut01: maybe we should set ```consume_preamble_events``` to ```True```

mbed_host_tests/host_tests_plugins/module_power_cycle_mbed.py
```diff
@@ -60,26 +60,28 @@ def execute(self, capability, *args, **kwargs):
             if capability in HostTestPluginPowerCycleResetMethod.capabilities:
                 target_id = kwargs['target_id']
                 device_info = kwargs['device_info']
-                result = self.__hw_reset(target_id, device_info)
+                ret = self.__get_mbed_tas_rm_addr()
+                if ret:
+                    ip = ret[0]
```
PrzemekWirkus:
```python
ip, port = ret
```

mbed_host_tests/host_tests_runner/host_test_default.py
mazimkhan: That will reload the host test. Remember host test manages the state for the target. Hence it should not be reloaded.

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -174,11 +174,11 @@ def conn_process(event_queue, dut_event_queue, prn_lock, config):
             break
 
         # Send data to DUT
-        if not dut_event_queue.empty():
-            try:
-                (key, value, _) = dut_event_queue.get(timeout=1)
-            except QueueEmpty:
-                continue
+        try:
+            (key, value, _) = dut_event_queue.get(timeout=0)
```
PrzemekWirkus: [method Queue.get(block, timeout)](https://docs.python.org/2/library/queue.html#Queue.Queue.get)
> Remove and return an item from the queue. If optional args block is true and timeout is None (the default), block if necessary until an item is available. **If timeout is a positive number, it blocks at most timeout seconds and raises the Empty exception if no item was available within that time**. Otherwise (block is false), return an item if one is immediately available, else raise the Empty exception (timeout is ignored in that case).

Are you sure setting timeout to 0 will raise `Empty` exception (`QueueEmpty` is alias for `Empty`).

mazimkhan: Yes. Just as it says. I think simpler way to explain could have been that either
block indefinitely if not timeout=None
or
block until a timeout or something to return.

Hence if not blocking and nothing to return then raise the exception.

PR #85, opened 2016-05-06
Add mount point checks with mbed_lstools before flashing

PrzemekWirkus:
## Description
We are not protecting test run from device mount point change when calling `mbedhtrun`.

## Changes:
* Add to `shell` and `shutil` plugins mount guards checking if mount point changed and/or is valid.
  * If mount point changed warning will be issues and new mount point will be used for copy/flash procedure. 
* Add `htrun` version print (first print by `htrun`) for debugging purposes:
```
[1462545824.12][HTST][INF] host test executor ver. 0.2.6
```

@mazimkhan @adbridge @bridadan Please review. 
This will be delivered separately to features:
* Serial port check before reset and
* HW reset recovery feature I think.

## Example usage
I've called `mbedhtrun` with correct `target_id` and wrong mount point `X:` and got correct response from `HostTestPluginCopyMethod_Shell` plugin with correct mount point value of mount point from `mbed_lstools`.

```
$ mbedhtrun -d X: -p COM218:9600 -f ".\build\frdm-k64f-gcc\test\mbed-drivers-test-generic_tests.bin" -C 4 -c shell -m K64F -t 0240000033514e450043500585d4003fe981000097969900
```
```
[1462545824.12][HTST][INF] host test executor ver. 0.2.6
[1462545824.12][HTST][INF] copy image onto target...
Plugin info: HostTestPluginCopyMethod_Shell::CopyMethod: Mount point for tid='0240000033514e450043500585d4003fe981000097969900' changed from 'X:' to 'E:'...
        1 file(s) copied.
[1462545832.54][HTST][INF] starting host test process...
```

### DIFF
--- a/mbed_host_tests/host_tests_plugins/host_test_plugins.py
+++ b/mbed_host_tests/host_tests_plugins/host_test_plugins.py
@@ -20,6 +20,7 @@ Author: Przemyslaw Wirkus <Przemyslaw.Wirkus@arm.com>
 import os
 import sys
 import platform
+import mbed_lstools
 
 from os import access, F_OK
 from sys import stdout
@@ -107,16 +108,36 @@ class HostTestPluginBase:
         stdout.flush()
         return True
 
-    def check_mount_point_ready(self, destination_disk, init_delay=0.2, loop_delay=0.25):
+    def check_mount_point_ready(self, destination_disk, init_delay=0.2, loop_delay=0.25, target_id=None):
         """! Waits until destination_disk is ready and can be accessed by e.g. copy commands
-
         @return True if mount point was ready in given time, False otherwise
-
         @param destination_disk Mount point (disk) which will be checked for readiness
         @param init_delay - Initial delay time before first access check
         @param loop_delay - polling delay for access check
         """
+
+        if target_id:
+            mbeds = mbed_lstools.create()
+            mbeds_by_tid = mbeds.list_mbeds_by_targetid()   # key: target_id, value mbedls_dict()
+
+            # Wait for mount point to appear with mbed-ls
+            # and if it does check if mount point for target_id changed
+            # If mount point changed, use new mount point and check if its ready (os.access)
+            new_destination_disk = destination_disk
+            for i in range(25): # 25x 200ms = 5sec
+                if target_id in mbeds_by_tid:
+                    if 'mount_point' in mbeds_by_tid[target_id]:
+                        new_destination_disk = mbeds_by_tid[target_id]['mount_point']
+                        break
+                sleep(200)
+
+            if new_destination_disk != destination_disk:
+                # Mount point changed, update to new mount point from mbed-ls
+                self.print_plugin_info("Mount point for tid='%s' changed from '%s' to '%s'..."% (target_id, destination_disk, new_destination_disk))
+                destination_disk = new_destination_disk
+
         result = False
+        # Check if mount point we've promoted to be valid one (by optional target_id check above)
         # Let's wait for 30 * loop_delay + init_delay max
         if not access(destination_disk, F_OK):
             self.print_plugin_info("Waiting for mount point '%s' to be ready..."% destination_disk, NL=False)
@@ -127,7 +148,7 @@ class HostTestPluginBase:
                     break
                 sleep(loop_delay)
                 self.print_plugin_char('.')
-        return result
+        return (result, destination_disk)
 
     def check_parameters(self, capability, *args, **kwargs):
         """! This function should be ran each time we call execute() to check if none of the required parameters is missing
--- a/mbed_host_tests/host_tests_plugins/host_test_plugins.py
+++ b/mbed_host_tests/host_tests_plugins/host_test_plugins.py
@@ -117,19 +117,24 @@ class HostTestPluginBase:
         """
 
         if target_id:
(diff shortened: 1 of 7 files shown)


mbed_host_tests/host_tests_plugins/host_test_plugins.py
```diff
@@ -107,16 +108,36 @@ def print_plugin_char(self, char):
...
+            # Wait for mount point to appear with mbed-ls
+            # and if it does check if mount point for target_id changed
+            # If mount point changed, use new mount point and check if its ready (os.access)
+            new_destination_disk = destination_disk
+            for i in range(25): # 25x 200ms = 5sec
+                if target_id in mbeds_by_tid:
+                    if 'mount_point' in mbeds_by_tid[target_id]:
+                        new_destination_disk = mbeds_by_tid[target_id]['mount_point']
+                        break
+                sleep(200)
```
bridadan: The unit of time for `sleep` is seconds, not milliseconds (https://docs.python.org/2/library/time.html#time.sleep). `200` should be `0.2`.

bridadan: Also, sometimes OSes take a long time to mount devices. I've seen delays up to ~10 seconds. I'd recommend extending this to about 30 seconds - 1 minute. Just probably print warnings for how long you're waiting so it doesn't look like it froze.

bridadan:
Instead of checking for the moint point to be ready in each copy plugin, maybe it should be moved to the parent function that calls the copy plugin? Like `copy_image` or `copy_image_raw`? (https://github.com/ARMmbed/htrun/blob/b71adf61457a0f1c47f8983f8cf2f3bb5af5243d/mbed_host_tests/host_tests_runner/mbed_base.py#L80)

mbed_host_tests/host_tests_plugins/host_test_plugins.py
mazimkhan: ```mbeds = mbed_lstools.create()``` should be done inside the loop. Otherwise it will loop on same data.
Like:
```
            mbeds = mbed_lstools.create()
            mbeds_by_tid = mbeds.list_mbeds_by_targetid()
            for i in range(25): # 25x 200ms = 5sec
                if target_id in mbeds_by_tid:
                    if 'mount_point' in mbeds_by_tid[target_id]:
                        new_destination_disk = mbeds_by_tid[target_id]['mount_point']
                        break
                sleep(to)
```

mbed_host_tests/host_tests_plugins/module_copy_mbed.py
```diff
@@ -72,6 +72,9 @@ def execute(self, capability, *args, **kwargs):
             self.print_plugin_error("Error: destination disk not specified")
             return False
 
+        # This optional parameter can be used if TargetID is provided (-t switch)
+        target_id = kwargs['target_id']
+
```
mazimkhan: If it is optional then we should check before accessing it. Like:
```
target_id = kwargs.get('target_id', None)
```

mbed_host_tests/host_tests_plugins/module_copy_shell.py
```diff
@@ -53,13 +53,18 @@ def execute(self, capability, *args, **kwargs):
             self.print_plugin_error("Error: destination disk not specified")
             return False
 
+        # This optional parameter can be used if TargetID is provided (-t switch)
+        target_id = kwargs['target_id']
```
mazimkhan: If it is optional then we should check before accessing it. Like:
```
target_id = kwargs.get('target_id', None)
```

PrzemekWirkus:
> Instead of checking for the moint point to be ready in each copy plugin, maybe it should be moved to the parent function that calls the copy plugin? Like copy_image or copy_image_raw? 

IMHO, It is a matter or choice I think. If user passes `target_id` it enabled for copy plugin to do dirty work and finds new mount point. No need to ask user to know other API to do so. Those two (mount point discovery/checks/flashing are not separate concerns).

PrzemekWirkus:
@bridadan @mazimkhan Thank you for great comments, please final check me :)

mbed_host_tests/host_tests_plugins/host_test_plugins.py
```diff
@@ -107,16 +108,41 @@ def print_plugin_char(self, char):
...
+            for i in range(120):
+                # mbed_lstools.create() should be done inside the loop.
+                # Otherwise it will loop on same data.
+                mbeds = mbed_lstools.create()
+                mbeds_by_tid = mbeds.list_mbeds_by_targetid()   # key: target_id, value mbedls_dict()
+                if target_id in mbeds_by_tid:
+                    if 'mount_point' in mbeds_by_tid[target_id]:
+                        new_destination_disk = mbeds_by_tid[target_id]['mount_point']
+                        break
+                self.print_plugin_info("Waiting for '%s' mount point (current is '%s')..."% (target_id, destination_disk))
```
bridadan: I don't think you'll want to print this every half a second. I think you'll want to print it once, with a warning saying that you'll wait for a minute before failing. Then proceed to check every half a second. Otherwise the log is gonna get pretty crazy :)

bridadan:
Other than my one new comment about the logging, looks good!

+1

PR #86, opened 2016-05-06
Add serial port checks with mbed_lstools before opening port by listener

PrzemekWirkus:
## Description
We are not protecting test run from device serial port change when calling `mbedhtrun`.

## Changes:
* Add to `mbed_reset` plugins guards checking if serial port changed and/or is valid.
  * If serial port changed warning will be issues and new serial port will be used for reset/sendBreak procedure.

@mazimkhan @adbridge @bridadan Please review. 
This will be delivered separately to features:
* Mount point check before reset and
* HW reset recovery feature I think.

## Example usage
I've called `mbedhtrun` with correct `target_id` and wrong serial port `COMx:` and got correct response from `HostTestPluginBase::BasePlugin` plugin with correct serial port name from `mbed_lstools`.

Note: Connection listener will open serial port and keep it open while later reset plugin will reuse OBJECT not PORT name so we only check port name before we open and keep serial port.

```
$ mbedhtrun -d E: -p COMx:9600 -f ".\build\frdm-k64f-gcc\test\mbed-drivers-test-generic_tests.bin" -C 4 -c shell -m K64F -t 0240000033514e450043500585d4003fe981000097969900
```
```
[1462549672.72][HTST][INF] copy image onto target...
        1 file(s) copied.
[1462549681.08][HTST][INF] starting host test process...
[1462549681.53][CONN][INF] starting connection process...
[1462549681.53][CONN][INF] initializing serial port listener...
[1462549681.53][SERI][INF] serial(port=COMx, baudrate=9600)
Plugin info: HostTestPluginBase::BasePlugin: Serial port for tid='0240000033514e450043500585d4003fe981000097969900' changed from 'COMx' to 'COM218'...
[1462549681.59][SERI][INF] serial port changed from 'COMx to 'COM218')
[1462549681.59][SERI][INF] reset device using 'default' plugin...
[1462549681.84][SERI][INF] wait for it...
[1462549682.84][CONN][INF] sending preamble '9cffc1bb-6a6c-4c4e-9fdc-350e7edb84d5'...
[1462549682.84][SERI][TXD] mbedmbedmbedmbedmbedmbedmbedmbedmbedmbed
[1462549682.84][SERI][TXD] {{__sync;9cffc1bb-6a6c-4c4e-9fdc-350e7edb84d5}}
[1462549682.97][CONN][INF] found SYNC in stream: {{__sync;9cffc1bb-6a6c-4c4e-9fdc-350e7edb84d5}}, queued...
[...]

PR #88, opened 2016-05-09
mbed-ls can return data for target id with wields like mount_point or serial_port set to None

PrzemekWirkus:
Changes:
* This patch makes sure we are not taking 'unknown' device property while pooling for it.
* Improvements for two previous PRs. See https://github.com/ARMmbed/htrun/pull/85 and https://github.com/ARMmbed/htrun/pull/86.


PrzemekWirkus:
@bridadan @mazimkhan This is small improvement for #85 and #86.

mazimkhan:
+1

bridadan:
+1

Issue #89, opened 2016-05-11
Serial port mbed-ls guard failed


PrzemekWirkus:
```
04:05:55 	calling mbedhtrun: mbedhtrun -d /mnt/DAPLINK -p /dev/ttyACM1:9600 -f "./build/frdm-k64f-gcc/test/configuration-store-test-find.bin" -C 4 -c shell -m K64F -t 0240000033514e450054500585d4001de981000097969900
04:05:55 mbedgt: mbed-host-test-runner: started
04:05:57 [1462935927.28][HTST][INF] host test executor ver. 0.2.8
04:05:57 [1462935927.28][HTST][INF] copy image onto target...
04:06:37 Plugin info: HostTestPluginCopyMethod_Shell::CopyMethod: Waiting for '0240000033514e450054500585d4001de981000097969900' mount point (current is '/mnt/DAPLINK')...
04:06:37 [1462935967.19][HTST][INF] starting host test process...
04:06:37 [1462935967.21][CONN][INF] starting connection process...
04:06:37 [1462935967.21][CONN][INF] initializing serial port listener...
04:06:37 [1462935967.22][SERI][INF] serial(port=/dev/ttyACM1, baudrate=9600)
04:06:44 [1462935974.08][SERI][ERR] [Errno 2] could not open port /dev/ttyACM1: [Errno 2] No such file or directory: '/dev/ttyACM1'
04:06:44 [1462935974.08][CONN][INF] sending preamble '40a38485-e62e-4c96-b8e7-bdddff50ba1f'...
04:06:44 [1462935974.08][SERI][TXD] {{__sync;40a38485-e62e-4c96-b8e7-bdddff50ba1f}}
04:06:44 [1462935974.08][HTST][ERR] connection lost, serial.Serial(/dev/ttyACM1. 9600, 0): [Errno 2] could not open port /dev/ttyACM1: [Errno 2] No such file or directory: '/dev/ttyACM1'
04:06:44 [1462935974.08][HTST][WRN] stopped to consume events due to __notify_conn_lost event
04:06:44 [1462935974.08][HTST][INF] test suite run finished after 6.89 sec...
04:06:44 [1462935974.09][HTST][INF] CONN exited with code: 0
04:06:44 [1462935974.09][HTST][INF] No events in queue
04:06:44 [1462935974.09][HTST][INF] host test result() call skipped, received: ioerr_serial
04:06:44 [1462935974.09][HTST][WRN] missing __exit event from DUT
[...]

PR #90, opened 2016-05-11
Add serial port mount time delay to 1 minute

PrzemekWirkus:
Note: there might be up to 3 minutes delay in mount on some OSs.

Changes:
* Change pooling for serial port to 1 minute (it will be a sec longeer because mbed-ls calls take extra few ms)
* Add simple "start pooling for serial" print for debuging.

Note: there can be following this one PR with actual pooling time switch command in mbedhtrun.

This PR tries to address https://github.com/ARMmbed/htrun/issues/89.


PrzemekWirkus:
@mazimkhan @bridadan @adbridge 
Small change so we wait for serial port longer (disk usually appears first and serial later).

bridadan:
+1

PR #97, opened 2016-06-12
Workaround for serial delays by never blocking on the dut_event_queue

### LINKED ISSUE #96: Serial communication introduces second long delays
Most noticable on echo tests. Causes the [echo tests](https://github.com/ARMmbed/mbed-os/blob/master/TESTS/mbed_drivers/echo/main.cpp) to fail without a timeout of >2 minutes.

cc @PrzemekWirkus @bridadan

geky:
Fixes #96 

I suspect there may be a better way to solve this, but this works as a temporary solution.


PrzemekWirkus:
@geky How this was tested?

geky:
Tested on Linux and Windows (not Mac)

PR #98, opened 2016-06-14
Add back support for -R <timeout-after-reset> command line switch

PrzemekWirkus:
Changes:
* Switch `-R` was not supported in new event loop `CONN` process.


PrzemekWirkus:
@bridadan @mazimkhan FYI

Issue #102, opened 2016-06-21
Remove custom logger from htrun


PrzemekWirkus:
# Description
Currently we are using [custom logger](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_conn_proxy/conn_proxy_logger.py) to log all events in `htrun` threads / processes we are using.
We want to replace custom logger which is using custom locking with python module logging.

# Rationale
* This change will reduce number of issues we have with extra code and misuse of locks.
* This change will help us to align with more generic solutions provided by Python standard modules.

# Definition of Done
* Replace [custom logger](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_conn_proxy/conn_proxy_logger.py) with [python logging module](https://docs.python.org/2/library/logging.html#module-logging) mechanism.
* Remove [custom logger](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_conn_proxy/conn_proxy_logger.py) implementation from source code completely.
* Refactor code:
  * Remove all references to locking used to manipulate custom logger.
* All unit tests must pass.
* Logging must be kept in the same format: timestamp, name, severity, text.
* Logging from few example test cases executions (e.g. `mbed-drivers`) must be virtually the same.

# Testing
* `htrun` must report the same line format with custom logger.
* Note that you can use `$python setup.py test` to run basic unit tests for `htrun`.

PR #103, opened 2016-06-22
Converted HtrunLogger to use Python 'logging' module removing custom lock

ConorPKeegan:
Adaption of the 'HtrunLogger' class to replace the custom outputting to stdout, with the builtin 'logging' module in Python. Now all of the logging will go through the 'logging' module, and will be formatted in the same way as before. Further iterations will fully remove the class, and rely solely on calling the logging code. The issue for this enhancement is #102.

@adbridge 
@mazimkhan 
@PrzemekWirkus


PrzemekWirkus:
+1

PR #111, opened 2016-07-15
Turn down the logging noise.

mjs-arm:
This set of patches taken together move the logging stream from stdout to stderr and turn the volume down slightly, defaulting to errors only.

We add a new option to allow the log level to be set on the command line.


PrzemekWirkus:
@mjs-arm Marcus, this change breaks interface between Greentea and htrun. If you suppress few prints you will get error results (I've checked, can paste here proof) in passing tests.
I think before we can merge this we need to identify prints that should be always present and give them the same level (always on). Rest can be suppressed.

Issue #115, opened 2016-07-25
[devel_framework_refactoring] NameError: global name 'result' is not defined


PrzemekWirkus:
# Description

When running all tests from mbed-os I've notices some issues inside (traceback). Please check.
Command to reproduce:
```
~/mbed-os$ mbedgt -V --report-html new03.html
```

# Traceback from `tests-mbed_drivers-rtc` test
I was running some tests with mbed-os and saw this issue in even loop:
```
Traceback (most recent call last): File "c:\Python27.11\Scripts\mbedhtrun-script.py", line 9, in
    load_entry_point('mbed-host-tests', 'console_scripts', 'mbedhtrun')()
  File "c:\work\pypi\htrun\mbed_host_tests\mbedhtrun.py", line 34, in main
    result = test_selector.execute()
  File "c:\work\pypi\htrun\mbed_host_tests\host_tests_runner\host_test_default.py", line 164, in execute
    test_result = self.run_test()
  File "c:\work\pypi\htrun\mbed_host_tests\host_tests_runner\host_test_default.py", line 133, in run_test
    return event_loop.run_loop()
  File "c:\work\pypi\htrun\mbed_host_tests\host_tests_event_loop\base_event_loop.py", line 88, in run_loop
    self.run_finish()
  File "c:\work\pypi\htrun\mbed_host_tests\host_tests_event_loop\default_event_loop.py", line 227, in run_finish
    self.logger.prn_inf("host test result(): %s"% str(result))
NameError: global name 'result' is not defined
```

See [HTML report](https://rawgit.com/PrzemekWirkus/4523d3a1c0d4f2f6a180a146f53ef988/raw/9fc6ba3072dadc3326046d7717763c36e795af1e/issue_115.html) for more details.

PrzemekWirkus:
@ConorPKeegan FYI, please check!

PR #116, opened 2016-07-25
Fix for Issue #115

ConorPKeegan:
# Changes
* Added missing `self` when using `self.result`

# Fixes
* Fixes issue #115


ConorPKeegan:
@PrzemekWirkus Here is a fix

PR #118, opened 2016-07-27
Fixed issue with resetting Conn Process

ConorPKeegan:
# Changes
* Moved the creation of conn process into event loop
* Fixed issue when a reset was needed the function to generate the conn process had been removed
* Removed some unneeded parameters and imports


ConorPKeegan:
@PrzemekWirkus

PrzemekWirkus:
I will check this tomorrow morning ! :)

PR #119, opened 2016-08-03
Assume test pass if no error reported.

mazimkhan:
On successful completion of binary htrun calls host test's ```result()``` method to get the result. Not all host tests implement this function. Also it is redundant. If host test detect a failure in mid of execution i.e. the scenario in which a host test sets ```result=False``` it can rather call ```notify_complete()```. For no error result can be assumed successful.

Also modified the host tests with result() function removed.

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -104,7 +104,7 @@ def print_ht_list(verbose=False):
     for column in column_names:
         pt.align[column] = 'l'
 
-    # Opiortunistic approach: always try to add current ./test/host_tests
+    # Oportunistic approach: always try to add current ./test/host_tests
     enum_host_tests('./test/host_tests', verbose=verbose)
 
     ht_str_len = 0
--- a/mbed_host_tests/host_tests_toolbox/host_functional.py
+++ b/mbed_host_tests/host_tests_toolbox/host_functional.py
@@ -81,11 +81,11 @@ def reset_dev(port=None,
             result = False
     return result
 
-def handle_send_break_cmd(port, 
-                          disk, 
-                          reset_type=None, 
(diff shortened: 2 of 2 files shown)


mazimkhan:
@bridadan @ConorPKeegan Please review

mbed_host_tests/host_tests/rtc_auto.py
```diff
@@ -34,14 +34,14 @@ def _callback_rtc(self, key, value, timestamp):
         self.rtc_reads.append((key, value, timestamp))
 
     def _callback_end(self, key, value, timestamp):
-        self.notify_complete()
+        self.notify_complete(self.varify())
```
ConorPKeegan: I think this should be `self.verify()` not `varify`

mazimkhan:
Can not merge yet as it is under testing and breaks some tests.

bridadan:
@mazimkhan I think this may be the actual cause of the failures we've been intermittently seeing: https://github.com/ARMmbed/htrun/pull/121

I'm currently testing this PR in CI at the moment, I'll let you know how it affects stability.

mazimkhan:
@bridadan any feedback. Should we merge this PR.

bridadan:
@mazimkhan I think assuming the test passed is a dangerous thing to do. I would probably assume `ERROR` to be safe (in case some unforeseen case crops ups). The last thing want to do is report a failing test is actually passing.

However, removing the `result` function might make sense here, allowing us to standardize on using `__notify_complete` instead. I admit I'm familiar with all of the side effects this has at the moment.

mazimkhan:
@bridadan I agree about PASS being default. Actually I saw it from one angle only. I will rework it.

PR #121, opened 2016-08-04
Fixing issue with multiprocess Queue

bridadan:
#### I will update this PR with results after running this in CI for sometime

This commit fixes an intermittent failure with the multiprocess Queue used
within the host test. There is a chance that a Queue will return that it
is empty even if an item has been added to it (see the Queue documentation
here:
https://docs.python.org/2/library/multiprocessing.html#pipes-and-queues).
This issue intermittently arrises in the following scenario:

- A DUT sends the {{end;success}} KV pair
- A DUT sends the {{__exit;0}} KV pair
- The {{end;success}} KV pair is pulled from the queue, which puts a
  callback for notify_complete to be called (after the __exit KV pair)
- The {{__exit;0}} KV pair is pulled from the queue, which breaks out of
  the loop
- The queue is checked to see if it is empty. This is the intermittent
  point of failure, as this sometimes returns True even though there
  should be a notify_compelte KV pair in the queue
- Since the queue is presumed to be empty, the result isn't properly
  updated and the test fails incorrectly

The fix for this is not to break out of the queue loop on __exit or on any
other KV pair or error condition. Instead, replace these 'break'
statements with another queue event called '__exit_event_queue'. This is
the only event that will call 'break'. The rule for this event is that **it
will always be the last event in the queue**. Since the Queue.empty()
function cannot be trusted, this is used instead to mark the end of the
host test execution.

This solution was inspired by the blog post here:
http://bryceboe.com/2011/01/28/the-python-multiprocessing-queue-and-large-objects/

cc @mazimkhan @ConorPKeegan

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ Key-value protocol was developed and is used to provide communication layer betw
       * ```__testcase_start``` - sent by DUT, test case start data.
       * ```__testcase_finish``` - sent by DUT, test case result.
       * ```__exit``` - sent by DUT, test suite execution finished.
+      * ```__exit_event_queue``` - sent by host test, indicating no more events expected.
   * Non-Reserved event/message keys have leading ```__``` in name:
     * ```__rxd_line``` - Event triggered when ```\n``` was found on DUT RXD channel. It can be overridden (```self.register_callback('__rxd_line', <callback_function>)```) and used by user. Event is sent by host test to notify a new line of text was received on RXD channel. ```__rxd_line``` event payload (value) in a line of text received from DUT over RXD.
 * Each host test (master side) has four functions used by async framework:
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -132,6 +132,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         callbacks_consume = True
         # Flag check if __exit event occurred
         callbacks__exit = False
+        # Flag check if __exit_event_queue event occurred
+        callbacks__exit_event_queue = False
         # Handle to dynamically loaded host test object
         self.test_supervisor = None
         # Version: greentea-client version from DUT
@@ -216,7 +218,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                                         print line
                                     self.logger.prn_inf("==== Traceback end ====")
                                     result = self.RESULT_ERROR
-                                    break
+                                    event_queue.put(('__exit_event_queue', 0, time()))
 
                                 self.logger.prn_inf("host test setup() call...")
                                 if self.test_supervisor.get_callbacks():
@@ -228,7 +230,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                             else:
                                 self.logger.prn_err("host test not detected: %s"% value)
                                 result = self.RESULT_ERROR
-                                break
+                                event_queue.put(('__exit_event_queue', 0, time()))
 
                             consume_preamble_events = False
                         elif key == '__sync':
@@ -240,6 +242,11 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                             self.logger.prn_wrn("stopped to consume events due to %s event"% key)
                             callbacks_consume = False
                             result = self.RESULT_IO_SERIAL
+                            event_queue.put(('__exit_event_queue', 0, time()))
+                        elif key == '__exit_event_queue':
+                            # This event is sent by the host test indicating no more events expected
+                            self.logger.prn_inf("%s received"% (key))
+                            callbacks__exit_event_queue = True
                             break
                         elif key.startswith('__'):
                             # Consume other system level events
@@ -252,7 +259,6 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                             # or if value is None, value will be retrieved from HostTest.result() method
                             self.logger.prn_inf("%s(%s)"% (key, str(value)))
                             result = value
-                            break
                         elif key == '__reset_dut':
                             # Disconnect to avoid connection lost event
                             dut_event_queue.put(('__host_test_finished', True, time()))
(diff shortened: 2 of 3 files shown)


mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -346,9 +361,12 @@ def start_conn_process():
...
 
-        #if not callbacks__exit and not result:
-        if not callbacks__exit and result is None:
-            self.logger.prn_err("missing __exit event from DUT and no result from host test, timeout...")
+        if not callbacks__exit_event_queue:
+            self.logger.prn_wrn("missing __exit_event_queue event from host test")
+
+        #if not callbacks__exit_event_queue and not result:
+        if not callbacks__exit_event_queue and result is None:
+            self.logger.prn_err("missing __exit_event_queue event from host test and no result from host test, timeout...")
```
sarahmarshy: This line is longer than 80 chars https://www.python.org/dev/peps/pep-0008/#maximum-line-length

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -366,7 +366,8 @@ def start_conn_process():
 
         #if not callbacks__exit_event_queue and not result:
         if not callbacks__exit_event_queue and result is None:
-            self.logger.prn_err("missing __exit_event_queue event from host test and no result from host test, timeout...")
+            self.logger.prn_err("missing __exit_event_queue event from " + \
+                "host test and no result from host test, timeout...")
```
sarahmarshy: :dancer:

bridadan:
@mazimkhan I think you had some concerns on this not handling failure cases correctly. Do you remember which cases these were? So far this patch seems to have helped quite a bit in CI.

README.md
```diff
@@ -250,6 +250,7 @@ Key-value protocol was developed and is used to provide communication layer betw
       * ```__testcase_start``` - sent by DUT, test case start data.
       * ```__testcase_finish``` - sent by DUT, test case result.
       * ```__exit``` - sent by DUT, test suite execution finished.
+      * ```__exit_event_queue``` - sent by host test, indicating no more events expected.
```
mazimkhan: Above documentation of ```__notify_complete``` is wrong. It is actually sent by the host test. Otherwise it will be redundant with ```__testcase_finish```. Also it does not matter as its purpose is to stop htrun event loop.

```__exit_event_queue``` is same as ```__notify_complete```.

File: mbed_host_tests/host_tests_runner/host_test_default.py [hunk 4]
```diff
@@ -279,11 +285,16 @@ def start_conn_process():
...
                             callbacks_consume = False
                             result = self.RESULT_IO_SERIAL
-                            break
+                            event_queue.put(('__exit_event_queue', 0, time()))
                         elif key == '__exit':
                             # This event is sent by DUT, test suite exited
                             self.logger.prn_inf("%s(%s)"% (key, str(value)))
                             callbacks__exit = True
+                            event_queue.put(('__exit_event_queue', 0, time()))
+                        elif key == '__exit_event_queue':
```
mazimkhan: This handle is same as handling of ```__notify_complete``` at line 257.

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -279,11 +285,16 @@ def start_conn_process():
                             self.logger.prn_wrn("stopped to consume events due to %s event"% key)
                             callbacks_consume = False
                             result = self.RESULT_IO_SERIAL
-                            break
+                            event_queue.put(('__exit_event_queue', 0, time()))
```
mazimkhan: How does it improve on break. If handling of ```__exit_event_queue``` just sets ```callbacks__exit_event_queue = True``` then that can be done here!

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -346,9 +361,13 @@ def start_conn_process():
...
-        #if not callbacks__exit and not result:
-        if not callbacks__exit and result is None:
-            self.logger.prn_err("missing __exit event from DUT and no result from host test, timeout...")
+        if not callbacks__exit_event_queue:
+            self.logger.prn_wrn("missing __exit_event_queue event from host test")
+
+        #if not callbacks__exit_event_queue and not result:
+        if not callbacks__exit_event_queue and result is None:
+            self.logger.prn_err("missing __exit_event_queue event from " + \
+                "host test and no result from host test, timeout...")
```
mazimkhan: Code clarity is important than meeting 80 char limit. I would review the log message rather than splitting it. May be: ```"Timeout as event `__exit_event_queue` missing"```

mazimkhan:
@bridadan On your PR documentation. ```__notify_complete``` is not a mandatory event. It is send by the host test to terminate the loop. i.e. scenarios where host test determines the verdict. I don't think this change is required. 
The reason for ```result``` being none and ```missing __exit event from DUT...``` error message is that base host test specifies the method ```result()``` for getting verdict from host test and it is called in the loop. But this method is not implemented by most of the host tests. Implementing a default ```result()``` in base classes can be erroneous. 
Hence in #119  this method and use is removed and host test verdict is assumed True if no error or exception is raised by the host test.

README.md
bridadan: @mazimkhan I did not change the `__notify_complete` documentation. This is good to know though, it sounds like I can probably remove the extra `__exit_event_queue` event and just use `__notify_complete`. I will have to look at this again.

mbed_host_tests/host_tests_runner/host_test_default.py
bridadan: The point here is that there is now only one exit point for the event queue, which is needed by the bug explained in this PR. This event is used to signal that the queue is "empty" since there is quirk with the `Queue.empty()` function that sometimes returns true erroneously. `callbacks__exit_event_queue` should only be set in one place, otherwise I'd have to replicate that in 3 other events.

mbed_host_tests/host_tests_runner/host_test_default.py
bridadan: This isn't true, `__notify_complete` also sets the result, which is the whole reason why this problem is so intermittent. There are multiple ways to get the result from the host test, which leads to different states of the queue at the point where `__exit` is processed.

mbed_host_tests/host_tests_runner/host_test_default.py
bridadan: Fine with me, I can make this change

mazimkhan:
Based on our discussion this PR is based on the analysis that ```___notify_complete``` is inserted after ```__exit```. This is possible because ```___notify_complete``` is inserted by same thread as the main loop and ```__exit``` is inserted by conn process. 
Second part of the analysis says that after exiting loop checking queue as ```while not event_queue.empty():``` may detect ```___notify_complete``` because of some known caveat in queue. 
Hence the solution is to resolve the contingency between ```__exit``` and ```__notify_complete```. This PR implements this by not exiting on ```__exit``` and delaying break by introducing new event ```__exit_event_queue```. Hence making behaviour deterministic.

LGTM

mazimkhan:
@bridadan Is it ready for merge or your still making some changes?

bridadan:
@mazimkhan Ready to go!

PR #122, opened 2016-08-23
Serial write timeout

bridadan:
Writing to a Serial port works differently on Linux than on Windows. The
pySerial documentation says that the `serial.write()` function will default
to blocking if `write_timeout` is not set. On Linux, this means that the
`serial.write()` blocks forever if the device has not received the data (which
happens in certain cases for different boards). This also occurs in
Windows, however it appears that Windows has a relatively large internal
buffer (~10K) that will cause the serial.write() function to return
immediately if there is still space in the buffer. Only when you send
large amounts of data does the function block.

This PR attempts to standardize the behavior of the host test (and to
prevent indefinite hangs) by introducing a default write timeout of 5
seconds.

It also adds checks throughout the `conn_proxy` to make sure the
call to `write()` actually succeeda. If not, it sets the error and exits.
This commit changes the return value of the `write()` function to `None` if
a write failure occurs. This also affects the function `write_kv()`.

cc @mazimkhan

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
@@ -124,7 +124,7 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
             self.selected_resource.write(payload)
             if log:
                 self.logger.prn_txd(payload)
-        return payload
+        return True
 
     def flush(self):
         pass
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
@@ -100,7 +100,7 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
         """! Read data from serial port RX buffer """
         # TIMEOUT: Since read is called in a loop, wait for self.timeout period before calling serial.read(). See
         # comment on serial.Serial() call above about timeout.
-        time.sleep(self.timeout)
+        time.sleep(self.read_timeout)
         c = str()
         try:
             if self.serial:
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
@@ -118,11 +118,11 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
                 self.serial.write(payload)
                 if log:
                     self.logger.prn_txd(payload)
+                return payload
         except SerialException as e:
             self.serial = None
             self.LAST_ERROR = "connection lost, serial.write(%d bytes): %s"% (len(payload), str(e))
             self.logger.prn_err(str(e))
-        return payload
 
     def flush(self):
         if self.serial:
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
@@ -29,15 +29,14 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
         ConnectorPrimitive.__init__(self, name)
         self.port = port
         self.baudrate = int(baudrate)
-        self.timeout = 0.01  # 10 milli sec
+        self.read_timeout = 0.01  # 10 milli sec
+        self.write_timeout = 5
         self.config = config
         self.target_id = self.config.get('target_id', None)
         self.polling_timeout = config.get('polling_timeout', 60)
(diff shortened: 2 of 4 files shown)


mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
```diff
@@ -101,11 +103,11 @@ def write(self, payload, log=False):
                 self.serial.write(payload)
                 if log:
                     self.logger.prn_txd(payload)
+                return payload
         except SerialException as e:
             self.serial = None
             self.LAST_ERROR = "connection lost, serial.write(%d bytes): %s"% (len(payload), str(e))
             self.logger.prn_err(str(e))
-        return payload
```
mazimkhan: In Python in absence of ```return``` statement LHS value will get a ```None```. Lets not rely on this implicit behaviour. It is less readable. and don't know if it may change!
Perhaps in exception block ```payload=None``` can be done leaving rest of the function intact.

mazimkhan: Also lets not return the same payload. As we either return the same payload or ```None```. Returning the payload perceived as written data makes the caller to put checks like ```if ret == data```. Though effectively it is boolean.
Hence return type can be bool. ```True``` if write is successful. ```False``` if write raises exception.

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -125,18 +125,33 @@ def conn_process(event_queue, dut_event_queue, config):
...
+        else:
+            return None
 
     # Send simple string to device to 'wake up' greentea-client k-v parser
-    connector.write("mbed" * 10, log=True)
+    wake_up_string = "mbed" * 10
+    written_wake_up_string = connector.write(wake_up_string, log=True)
+
+    # Failed to write 'wake up' string, exit conn_process
+    if written_wake_up_string != wake_up_string:
```
mazimkhan: Check suggestion above in ```write()``` to change ```write()``` return value to be boolean.

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -151,25 +166,36 @@ def __send_sync(timeout=None):
     if sync_behavior > 0:
         # Sending up to 'n' __sync packets
         logger.prn_inf("sending up to %s __sync packets (specified with --sync=%s)"% (sync_behavior, sync_behavior))
-        sync_uuid_list.append(__send_sync())
-        sync_behavior -= 1
+        sync_uuid = __send_sync()
```
mazimkhan: Can these two syncs be put in a loop? Making it configurable in future.

bridadan: @mazimkhan Maybe we can do that in another PR?

mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
bridadan: Done!

bridadan:
@mazimkhan I made the change you requested about the write function. Also rebased!

mazimkhan:
A minor comment here https://github.com/ARMmbed/htrun/pull/122/files#diff-ab29c0280949087d36c3c6cca0af4441R127

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -124,7 +124,7 @@ def write(self, payload, log=False):
             self.selected_resource.write(payload)
             if log:
                 self.logger.prn_txd(payload)
-        return payload
+        return True
```
bridadan: I wasn't really sure for the remote primitive. In the past the `payload` was always returned from write, which is why I just replaced it with True. I think we could probably revisit the remote primitive when we start seeing more of it's use.

mazimkhan: Ok no problem. Please rebase with master. Sorry I just merged the other PR #130. You may have to fix merge issues.

mazimkhan:
@bridadan  can you please merge with master HEAD.

mazimkhan:
@bridadan is it still needed?

bridadan:
Hi @mazimkhan, sorry your previous request fell through the cracks. This was a bit nontrivial to rebase, so I'm going to test on my local machine for a little bit before pushing up the changes. Thanks for prompting me :+1:

bridadan:
OK, I've ran through a couple of test runs locally, trying a few failure modes and things seem to be ok. Ready for re-review.

PR #123, opened 2016-08-29
Handle carriage returns if present

karlj:
I was running htrun on Linux Mint 17.3 and even though only new line characters were being written from the DUT, htrun was reading in a carriage return character and new line character. I added an optional carriage return character to the regex that searches for KV tokens.


mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -28,7 +28,7 @@
 class KiViBufferWalker():
     """! Simple auxiliary class used to walk through a buffer and search for KV tokens """
     def __init__(self):
-        self.KIVI_REGEX = r"\{\{([\w\d_-]+);([^\}]+)\}\}\n"
```
mazimkhan: Since the key value protocol is based on start ```{{``` and end ```}}`` tags. I would suggest removing ```\n```. That way it will be more robust to match. Even if on some platform ```\n``` goes missing.

karlj: I thought about doing that, but https://github.com/ARMmbed/htrun/pull/104 put the newline character in.

mazimkhan: @bridadan in #104 your have add ```\n``` to the K,V regex. I think K,V protocol and ```\n``` and ```\r``` are different things. If htrun should send after receiving a ```\n``` then we should received until ```\n``` and then do the parsing. So that K,V parsing is independent of these terminating characters and also since data in K,V has nothing to do with serial limitation. In future this protocol can be on network interface as well.

So the work around for preventing DAPLINK bad state should be in the place where we are receiving data i.e. receive until ```\n```.

bridadan: @mazimkhan Could you clarify a bit? I'm not sure what you mean. Are you saying you want to completely remove `\n` and/or `\r` from the KV regex?


The `\n` had been there before, but at when point @PrzemekWirkus had removed it, which caused issues in DAPLink, so I just added it back in that PR.

mazimkhan: Yes. It can be removed from the regex. But used after receiving characters to see if one line is received. To satisfy DAPlink limitation.

mazimkhan: @karlj @bridadan created PR #128 for making regex ```\n``` free.

mazimkhan:
closing as fix added in #128

mazimkhan:
closing as fix added in #128

Issue #87, opened 2016-05-08
Command options and the default behaviors


bridadan:
Pull requests #85 and #86 add support for checking the mount point and serial port for a platform. Since this information can be derived from mbed-ls (when supplied a target ID), it seems like the only time you would provide the serial port or mount point is when you don't know the target ID, the platform doesn't have a valid target ID, or you want to override one or many of the values that were resolved through mbed-ls.

If no target ID is provided, then it should default to requiring a provided serial port, baud rate, and mount point. If a target ID is provided, then you should be able to override the serial port, baud rate, and mount point through command line arguments.

It also seems valid to say sometimes you would want to override the baud rate without providing a serial port as part of the `-p` option (especially if you don't know the serial port). I would suggest adding a `--baud-rate` options to specify the serial port's baud rate specifically. The `-p` can still accept a baud rate component (`<port>:<baud rate>`) for backward compatibility. In the case that `-p <port>:<baud rate> --baud-rate <different baud rate>` are both specified, then `--baud-rate` would be used.

The priority for platform options from **lowest to highest**:
1. Values determined from target ID and mbed ls (with baud rate defaulting to 115200 or 9600)
2. `-d,--disk-path` for disk/mount point, `-p <port>:<baud rate>` for serial port (`:<baud rate>` bit being optional)
3. `--baud-rate` for serial baud rate

ciarmcom:
ARM Internal Ref: IOTSFW-2524

PrzemekWirkus:
Yes, this makes perfcet sense, we can iterate and change how -t behaves.

+1

mazimkhan:
@bridadan this issue is fixed here. We would need changes in Greentea to make use of this feature.

theotherjimmy:
@bridadan please reopen if this is still a problem.

PR #124, opened 2016-09-05
Add command line switch to specify baud rate

ConorPKeegan:
# Changes
* Add command line switch `--baud-rate` to specify the baud rate of the target
* Switch overrides value from target ID, mbed-ls, `-d,--disk-path` disk/mount point, `-p <port>:<baud rate>` for serial port

# Implements
* This pull request implements part of #87

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -58,7 +58,7 @@ HOSTREGISTRY.register_host_test("wait_us_auto", WaitusTest())
 HOSTREGISTRY.register_host_test("dev_null_auto", DevNullTest())
 
 # Set the default baud rate
-DEFAULT_BAUD_RATE = 115200
+DEFAULT_BAUD_RATE = 9600
 
 ###############################################################################
 # Functional interface for test supervisor registry
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -57,6 +57,9 @@ HOSTREGISTRY.register_host_test("default_auto", DefaultAuto())
 HOSTREGISTRY.register_host_test("wait_us_auto", WaitusTest())
 HOSTREGISTRY.register_host_test("dev_null_auto", DevNullTest())
 
+# Set the default baud rate
+DEFAULT_BAUD_RATE = 115200
+
 ###############################################################################
 # Functional interface for test supervisor registry
 ###############################################################################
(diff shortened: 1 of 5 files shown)


ConorPKeegan:
@mazimkhan @adbridge @bridadan Please can you review this

mbed_host_tests/host_tests_toolbox/host_functional.py
```diff
@@ -97,13 +97,16 @@ def handle_send_break_cmd(port,
...
+        baudrate = int(port_config[1]) if not baudrate else baudrate
     elif len(port_config) == 3:
         # -p COM4:115200:0.5
         port = port_config[0]
-        baudrate = int(port_config[1])
+        baudrate = int(port_config[1]) if not baudrate else baudrate
         timeout = float(port_config[2])
 
+    if not baudrate:
+        baudrate = 115200
```
mazimkhan: Can this be defined somewhere single place. Like ```mbed_host_tests.__init__.py```

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -60,6 +60,8 @@ def __init__(self, options):
             self.serial_baud = int(port_config[1])
             self.serial_timeout = float(port_config[2])
 
+        self.serial_baud = self.options.baud_rate if self.options.baud_rate else self.serial_baud
```
mazimkhan: Please put a comment here saying "Override baud rate with user supplied value" or something like that.

mbed_host_tests/host_tests_toolbox/host_functional.py
ConorPKeegan: @mazimkhan I'm not sure where I would put it in `mbed_host_tests.__init__.py`, as there are no constants defined there

mbed_host_tests/__init__.py
```diff
@@ -57,6 +57,9 @@
 HOSTREGISTRY.register_host_test("wait_us_auto", WaitusTest())
 HOSTREGISTRY.register_host_test("dev_null_auto", DevNullTest())
 
+# Set the default baud rate
+DEFAULT_BAUD_RATE = 115200
```
bridadan: I realize this default values are usually a heated topic, so I don't want to beat this to death. But all targets in mbed OS 5 default to a baud rate of 9600. Perhaps we should reflect that here as well since this is mainly used with mbed OS? Is there another reason (besides preference, speed, or that's what it was before) that this is 115200?

ConorPKeegan: @bridadan 115200 was the previous value, but I agree that it should probably be changed to 9600. @mazimkhan?

mazimkhan: Agree with Brian as its most common.

bridadan:
LGTM!

mazimkhan:
LGTM

PR #125, opened 2016-09-08
Add idle timeout for coverage

ConorPKeegan:
# Changes
* Detect when coverage is being output, and switch to an idle timeout, where the timer is reset whenever a new coverage packet is received.
* The idle timeout with timeout when a key value pair hasn't been received within the idle timeout duration, which by default is 10 seconds.
* Idle timeout saves the remaining time in the original timeout, and when a key that is not coverage is received, that timeout is restored.
* This is needed as coverage takes far longer to receive from the target than the timeouts that are set by the test, which causes most of the coverage data to be missed, and most of the tests to fail because of timeouts.

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -116,6 +116,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         """
         result = None
         timeout_duration = 10       # Default test case timeout
+        coverage_idle_timeout = 10  # Default coverage idle timeout
         event_queue = Queue()       # Events from DUT to host
         dut_event_queue = Queue()   # Events from host to DUT {k;v}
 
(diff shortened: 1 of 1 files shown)


ConorPKeegan:
@mazimkhan @bridadan Please can you review this

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -247,6 +251,29 @@ def start_conn_process():
...
                             self.logger.prn_err("orphan event in preamble phase: {{%s;%s}}, timestamp=%f"% (key, str(value), timestamp))
                     else:
+                        # If coverage detected use idle loop
+                        if key == '__coverage_start':
+                            if not coverage_idle:
+                                self.logger.prn_inf("starting coverage idle timeout loop...")
+                                coverage_idle = True
+
+                                # Save position in timeout to resume after coverage
+                                resume_timeout = timeout_duration - (time() - start_time)
```
mazimkhan: As discussed changing loop variables within the loop can be difficult to understand by some. It would be a good Idea to branch out into a different function that has the other loop driven by idle timeout for processing coverage data. After returning from that function the time taken by that function can be added to ```timeout_duration``` to accommodate time added by coverage data.

mazimkhan:
LGTM

PR #126, opened 2016-09-13
Increasing initial timeout to 1 minute

bridadan:
If a system is under heavy load, it can sometimes take longer than 10
seconds for the conn_process to start. This commit increases that timeout
to improve reliability on these systems.

This behavior can be seen in the following log from CI:

```
[1473742292.89][HTST][INF] host test executor ver. 1.1.0
[1473742292.89][HTST][INF] copy image onto target...
[1473742292.89][COPY][INF] Waiting up to 60 sec for '110100004420312033504139333031313431303497969903' mount point (current is 'L:')...
        1 file(s) copied.
[1473742305.41][HTST][INF] starting host test process...
[1473742315.41][HTST][INF] test suite run finished after 10.00 sec...
[1473742330.12][CONN][INF] starting connection process...
[1473742330.12][CONN][INF] notify event queue about extra 60 sec timeout for serial port pooling
[1473742330.12][CONN][INF] initializing serial port listener... 
[1473742330.12][SERI][INF] serial(port=COM13, baudrate=9600, timeout=0, write_timeout=5)
[1473742330.12][PLGN][INF] Waiting up to 60 sec for '110100004420312033504139333031313431303497969903' serial port (current is 'COM13')...
[1473742330.18][SERI][INF] reset device using 'default' plugin...
[1473742330.43][SERI][INF] waiting 1.00 sec after reset
[1473742331.43][SERI][INF] wait for it...
[1473742331.43][SERI][TXD] mbedmbedmbedmbedmbedmbedmbedmbedmbedmbed
[1473742331.43][CONN][INF] sending up to 2 __sync packets (specified with --sync=2)
[1473742331.43][CONN][INF] sending preamble '4abff833-32f6-4228-8073-b62f5e5539b2'
[1473742331.43][SERI][TXD] {{__sync;4abff833-32f6-4228-8073-b62f5e5539b2}}
[1473742331.43][CONN][INF] received special even '__host_test_finished' value='True', finishing
[1473742331.45][HTST][INF] CONN exited with code: 0
[1473742331.45][HTST][INF] Some events in queue
[1473742331.45][HTST][INF] stopped consuming events
[1473742331.45][HTST][INF] host test result(): None
[1473742331.45][HTST][WRN] missing __exit event from DUT
[1473742331.45][HTST][WRN] missing __exit_event_queue event from host test
[1473742331.45][HTST][ERR] missing __exit_event_queue event from host test and no result from host test, timeout...
[1473742331.45][HTST][INF] calling blocking teardown()
[1473742331.45][HTST][INF] teardown() finished
[1473742331.45][HTST][INF] {{result;timeout}}
```

You can see between the following lines:

```
[1473742305.41][HTST][INF] starting host test process...
[1473742315.41][HTST][INF] test suite run finished after 10.00 sec...
[1473742330.12][CONN][INF] starting connection process...
```

that it took 25 seconds to start the connection process thread. If the system is under heavy load, this may be reasonable, especially if `htrun` is always waiting for device to remount.

Please review @mazimkhan
FYI @adbridge

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -116,7 +116,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         @return Return self.TestResults.RESULT_* enum
         """
         result = None
-        timeout_duration = 10       # Default test case timeout
+        timeout_duration = 60       # Default test case timeout
         event_queue = Queue()       # Events from DUT to host
         dut_event_queue = Queue()   # Events from host to DUT {k;v}
 


mazimkhan:
The timeout is essentially for serial comm i.e. host test should be able to exchange ```__sync``` and host test script name and **target test timeout** within this initial timeout period. This initial timeout is impacted by the host machine performance i.e. the fork of conn process takes time. Can we not separate this fork of con process and serial comm timeout. So that we don't have to keep tuning the timeout. 

Reason for saying this is that increasing overall init/default timeout has an impact on test suite execution if a particular build is bad and every test fails to respond. Although it is just worst case scenario.

If we can wait for a new event ```__con_proc_started``` first from the started conn process then we can start the main loop after that i.e. not consuming conn process start time (that varies on machines) from the init timeout of 10 sec. 
@adbridge @bridadan what do you say?

Still I am ok with the current fix. but it is important to know actual reason and impact of the change.

mazimkhan:
This PR https://github.com/ARMmbed/htrun/pull/127 may improve performance on RPi.

bridadan:
@mazimkhan I see what you're saying about separating the timeout times. I will look into changing this.

bridadan:
@mazimkhan Currently the timeout for the conn_process isn't actually a timeout, since the process actually hangs even after the initial 10 seconds have passed. It's not until the process is started, a `__host_test_finished` is inserted into the process's queue, and the process ends gracefully that htrun actually exists.

How would you feel about making htrun not start the "timeout counter" until the conn process has actually started? This could be accomplished with a kv pair that is put by the conn process into the queue. Maybe something like `__conn_process_start`?

mazimkhan:
@bridadan that is exactly what I am proposing. But there should still be a timeout. Maybe for this it could be process fork timeout.

bridadan:
@mazimkhan I'm not sure its possible to set a timeout for starting a process in python. I did a quick check of the docs and I didn't see anything.

Could we still exit cleanly from the main process if we don't wait for the child process to start up?

I don't see much use in telling a child process to stop immediately after it starts (as is the case now if the child process takes up too long). It seems like the most useful thing is to give the process all the time it needs to start, then start the timer for timeout. The one time we would actually stop the child process from starting would be if someone Ctrl-C's/kills the parent process.

mazimkhan:
@bridadan I meant that wait for ```__conn_proc_start``` with ```event_queue.get(timeout=<proc start timeout>)```. This will ensure that we don't sit for ever if there is some issue in the process to start at all.

bridadan:
@mazimkhan I see what you're saying. I'm concerned about leaking a rouge child process. Do you know if you kill the parent process that all child processes will be killed as well?

mazimkhan:
@bridadan Sorry I don't understand how observing conn process start under a timeout would leak process. Can you not do ```Process.terminate()``` on timeout expiry. If this doesn't help then there is nothing you can do. You code can look like:
```
                    p = start_conn_process()
                    try:
                        (key, value, timestamp) = event_queue.get(timeout=60)
                        if key != '__conn_proc_start':
                            # gracefully terminate conn process
                            p = None
                    except QueueEmpty:
                        # log that couldn't start conn process
                        p.terminate()
                        p = None

                    # may be wrap the event loop inside a if
                    if p:
                        ....
```

Without a timeout there is chance that both conn process and htrun can get stuck.

bridadan:
@mazimkhan I guess my worry was because we don't know for sure if the Process has started correctly, I was wondering if terminating it would actually work. But I think it'll probably be fine. I will close this PR and submit a revised implementation.

PR #127, opened 2016-09-14
Fixed tight loop while checking event_queue

mazimkhan:
In actual code below there is a tight loop:
```
            while (time() - start_time) < timeout_duration:
                # Handle default events like timeout, host_test_name, ...
                self.logger.prn_inf("Tight loop")
                if not event_queue.empty():
                    try:
                        (key, value, timestamp) = event_queue.get(timeout=1)
                    except QueueEmpty:
                        continue    
```

```self.logger.prn_inf("Tight loop")``` line is added to demonstrate this. Try this and you can observe how much it is printed.
It can have a vicious impact on slower platforms like RPi where starting conn process takes time hence more spinning on checking ```queue.empty()```.

Check ```if not event_queue.empty():``` causes the tight loop as it evaluates to False until there is data in the queue. This check is redundant as check element in the queue inside a ```try``` block and catch ```QueueEmpty``` exception and return to the loop. ```try``` block is better as ```event_queue.get(timeout=1)``` waits ```timeout``` seconds before returning and does not cause uncontrolled spinning.

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -182,126 +182,125 @@ class DefaultTestSelector(DefaultTestSelectorBase):
             consume_preamble_events = True
             while (time() - start_time) < timeout_duration:
                 # Handle default events like timeout, host_test_name, ...
-                if not event_queue.empty():
-                    try:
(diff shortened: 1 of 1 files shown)


mazimkhan:
@bridadan @adbridge please review.

PR #128, opened 2016-09-14
KV parsing rework with newline termination in mind

mazimkhan:
Key,Value parsing is reworked to separate key,Value protocol format and need to limit request response based on ```\n```. See #104 for context. Also affects #123 

Additionally the ```self.buff``` was being appended with received strings without being cleared and without being needed as only ```self.buff_idx``` onward string was parsed. Looks like a memory leak.

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -28,26 +28,32 @@ from conn_primitive_remote import RemoteConnectorPrimitive
 class KiViBufferWalker():
     """! Simple auxiliary class used to walk through a buffer and search for KV tokens """
     def __init__(self):
-        self.KIVI_REGEX = r"\{\{([\w\d_-]+);([^\}]+)\}\}\n"
+        self.KIVI_REGEX = r"\{\{([\w\d_-]+);([^\}]+)\}\}"
         self.buff = str()
-        self.buff_idx = 0
+        self.kvl = []
         self.re_kv = re.compile(self.KIVI_REGEX)
 
     def append(self, payload):
-        """! Append stream buffer with payload """
+        """! Append stream buffer with payload and process"""
         self.buff += payload
+        lines = self.buff.split('\n')
+        self.buff = lines[-1]   # remaining
+        lines.pop(-1)
+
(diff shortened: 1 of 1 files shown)


mazimkhan:
@bridadan @adbridge

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -28,26 +28,32 @@
...
         self.buff = str()
-        self.buff_idx = 0
+        self.kvl = []
         self.re_kv = re.compile(self.KIVI_REGEX)
 
     def append(self, payload):
-        """! Append stream buffer with payload """
+        """! Append stream buffer with payload and process"""
         self.buff += payload
+        lines = self.buff.split('\n')
```
bridadan: Since we're now explicitly adding '\r' in https://github.com/ARMmbed/mbed-os/pull/2702, is there a place where this character can be handled and stripped in this PR?

Ideally the next release of htrun should be able to handle cases where the '\r' is or is not present, since the tool may be used with older releases of mbed OS.

bridadan: It'd probably fine just to do a `rstrip` on the returned line before inserting it into the `kvl` list: https://docs.python.org/2/library/string.html#string.rstrip

bridadan:
Changes seem ok to me besides my comments. The main thing I thing we have to be careful of with this PR is not breaking compatibility with older releases

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
mazimkhan: Since ```\n``` is removed from K,V regex any character outside the K.V pattern does not affect us. I have tested it. It works with mbed-os master and ARMmbed/mbed-os#2702

mazimkhan: Since we extract key and value using the K,V regex we don't need any extra processing like ```rstrip```.

bridadan: Ah I see @mazimkhan, thanks for explaining. Totally missed that the first time around :smile:

PR #130, opened 2016-09-21
Fixed performance issue due to spinning on non blocking Serial.read()

mazimkhan:
Three changes:
- Added a delay in call to ```Serial.read()``` as ```timeout``` is set to 0 making ```read()``` non blocking. This makes connection process loop to keep spinning on ```read()``` when there is no data resulting in high CPU usage. Serial timeout is not used as blocking ```read``` call impacts thread switching in python causing undefined behaviour. 
- Avoiding K,V pair lines/ data from printing as ```printf``` prints (```__rxd_line```) from the target.
- Removed use of ```event_queue.empty()``` from ```process_code_coverage()``` as it causes spinning in the loop when there is not element in queue.

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
@@ -29,7 +29,7 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
         ConnectorPrimitive.__init__(self, name)
         self.port = port
         self.baudrate = int(baudrate)
-        self.timeout = 0
+        self.timeout = 0.2  # 200 milli sec
         self.config = config
         self.target_id = self.config.get('target_id', None)
         self.serial_pooling = config.get('serial_pooling', 60)
@@ -51,7 +51,10 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
             self.port = serial_port
 
         try:
-            self.serial = Serial(self.port, baudrate=self.baudrate, timeout=self.timeout)
+            # TIMEOUT: While creating Serial object timeout is delibrately passed as 0. Because blocking in Serial.read
+            # impacts thread and mutliprocess functioning in Python. Hence, instead in self.read() s delay (sleep()) is
+            # inserted to let serial buffer collect data and avoid spinning on non blocking read().
+            self.serial = Serial(self.port, baudrate=self.baudrate, timeout=0)
         except SerialException as e:
             self.serial = None
             self.LAST_ERROR = "connection lost, serial.Serial(%s. %d, %d): %s"% (self.port,
@@ -84,6 +87,9 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
 
     def read(self, count):
         """! Read data from serial port RX buffer """
+        # TIMEOUT: Since read is called in a loop, wait for self.timeout period before calling serial.read(). See
+        # comment on serial.Serial() call above about timeout.
+        sleep(self.timeout)
         c = str()
         try:
             if self.serial:
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
@@ -29,7 +29,7 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
         ConnectorPrimitive.__init__(self, name)
         self.port = port
         self.baudrate = int(baudrate)
-        self.timeout = 0.2  # 200 milli sec
+        self.timeout = 0.01  # 10 milli sec
         self.config = config
         self.target_id = self.config.get('target_id', None)
         self.serial_pooling = config.get('serial_pooling', 60)
@@ -57,7 +57,7 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
             self.serial = Serial(self.port, baudrate=self.baudrate, timeout=0)
         except SerialException as e:
             self.serial = None
-            self.LAST_ERROR = "connection lost, serial.Serial(%s. %d, %d): %s"% (self.port,
+            self.LAST_ERROR = "connection lost, serial.Serial(%s, %d, %d): %s"% (self.port,
                 self.baudrate,
                 self.timeout,
                 str(e))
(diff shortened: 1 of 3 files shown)


mazimkhan:
@bridadan @adbridge Please review.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
```diff
@@ -84,6 +87,9 @@ def reset_dev_via_serial(self, delay=1):
 
     def read(self, count):
         """! Read data from serial port RX buffer """
+        # TIMEOUT: Since read is called in a loop, wait for self.timeout period before calling serial.read(). See
+        # comment on serial.Serial() call above about timeout.
+        sleep(self.timeout)
```
bridadan: Won't this have a pretty bad impact on timing tests? Now we only have a granularity of 200ms in which we can assign timestamps, which will cause significant timing drifts. I'm not sure we want to add this call to `sleep` quite yet.

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -194,23 +194,22 @@ def process_code_coverage(key, value, timestamp):
...
-                        continue
-
-                    # If coverage detected use idle loop
-                    # Prevent breaking idle loop for __rxd_line (occurs between keys)
-                    if key == '__coverage_start' or key == '__rxd_line':
-                        start_time = time()
+                try:
+                    (key, value, timestamp) = event_queue.get(timeout=1)
+                except QueueEmpty:
+                    continue
```
bridadan: Awesome optimization here!

mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
mazimkhan: Based on the relation between baud rate and bytes per sec for maximum baud rate of 115200 the max data rate would be ```11520``` Bps. For ```0.2``` sec receive window it will be ```2304``` bytes.  This tells we can only receive maximum ```2304``` bytes every ```0.2``` seconds. Hence if we have fear of drifting then we just have to increase our receive buffer to ```2304``` or more and pledge to process that before next ```0.2``` sec tick.

bridadan: @mazimkhan I'm not so much worried about data throughput, more data timing.

If the serial data comes in at any point while the conn_process is sleeping, we will lose up to 200ms of timing resolution for that test.

mazimkhan: Exactly since the timer tests we have are for 1 sec granularity they will not be impacted. For anything less then 1 sec host test based verification is not perfect anyways.

bridadan: But they will be impacted. Yes we print every 1 sec, but the host tests are constantly looking for drift between each tick. And the `sleep` call will prevent us from assigning a time stamp that is close to when we actually receive the serial data. This is the line I think that will be impacted: https://github.com/ARMmbed/htrun/pull/130/files#diff-9e6e1b40e765e6a2609375222e35117dR48

I agree that running timing tests on non-realtime  OS is not all that great, I was talking to @sg- about this. But at the moment will still do have these tests.

bridadan: Ok, I think I see what you're saying. Since all data is being shifted by that amount, it won't really show up as a failure. It may mask some drifting issues, but I think you're right that testing timing is already imperfect on a non-realtime host OS.

Do you think this change merits a note in the docs that serial data will only be processed ~5 times a second?

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -34,17 +34,31 @@ def __init__(self):
...
+                before = line[:pos]
+                after = line[pos + len(match):]
+                if len(before) > 0:
+                    discarded.append(before)
+                if len(after) > 0:
+                    discarded.append(after)
+            else:
+                # not a K,V pair
+                discarded.append(line)
+        return discarded
```
adbridge: I think some comments about what exactly this is doing would be useful here

mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
sg-: Can we determine if the drift measurement is needed with the CPU being relaxed? Given we only saw these issues under heavy parallelization in CI test farms there may be relationship here.

mazimkhan: @sg- buffering of data is done here https://github.com/mazimkhan/htrun/blob/80d13359cef7485bb2b2bc01fd6c9cdc332e5244/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py#L36
For pyserial ```timeout=0``` means non blocking i.e. ```read()``` returns immediately. For non-blocking it has to be set as ```None```. 
Based on the baud rate (max 115200) the data speed is limited and the timeout is short enough. Reducing timeout means aggressive polling for data. Since the data rate is already limited by the baud rate we don't need aggressive polling. But yes shorter timeout may help in preciseness in timing tests.

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
mazimkhan: Comments added

adbridge:
+1 LGTM

bridadan:
@mazimkhan The CPU usage is definitely way less! I can see there is a bit of a side effect with adding the sleep to the read. Basically all the times of the ticks align on that 200ms interval.

As a compromise, could we change the read delay from 200ms to 1ms? I ran this and tracked both options with Windows Task Manager:

**200ms**
200_ms_delay

**1ms**
1_ms_delay


As you can, see the CPU usage is pretty close to being the same. There's a little bit of a peak on CPU1 for the 1ms config, but I would think this is negligible over a long run. And we can always tweak the timing later.

mazimkhan:
@bridadan Can you please test this with parallel testing?

mazimkhan:
In process explorer it does seems to take 1% CPU 
image

mazimkhan:
With 200ms it is way <1%

image

bridadan:
Meet me in the middle somewhere, maybe 10ms? I'm trying to preserve some granularity.

mazimkhan:
Deal. I make it 10 ms.

bridadan:
@mazimkhan I'm testing this PR with parallel testing on Pis and a Windows machine and I've upped the timeout to 10ms. I'll keep you posted! Already seeing a little speed improvement!

mazimkhan:
Tested with ```10 ms``` much better. Merging.

PR #132, opened 2016-10-20
Check if mbed disconnects and connects back after flashing and before testing

mazimkhan:
It is observed in CI that opening serial port occasionally fails after flashing with error:

```
[1476968928.23][SERI][ERR] Cannot configure port, something went wrong. Original message: WindowsError(31, 'A device attached to the system is not functioning.')
```

This is suspected to be because copy command has finished and `htrun` has proceeded to opening the serial port. But board is still being programmed or is being remounted. Hence it is required we check that board is remounted after flashing and before starting testing.

Changes here check that the board is disconnected and connected back. This is done by obtaining a file handle in mbed mount point and waiting until it gets invalid indicating disconnect. Then waiting for mbed to connect back.

mbed-ls polling can't be used because it checks mbed at the instance it is run. Hence can't determine remount if mbed goes and comes back too quickly.  

This check for remount is only to enhance robustness. If remount does not happen no exception is raised.

### DIFF
--- a/mbed_host_tests/host_tests_runner/mbed_base.py
+++ b/mbed_host_tests/host_tests_runner/mbed_base.py
@@ -149,6 +149,7 @@ class Mbed:
                     if file.lower() == 'assert.txt':
                         assert_file = os.path.join(mbed_drive, file)
             if assert_file:
+                print "MBED: Removing %s" % assert_file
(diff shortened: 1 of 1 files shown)


mazimkhan:
Closing as it was for testing only.

PR #138, opened 2016-11-24
Polling for serial port to get ready (in linux)

mazimkhan:
On Linux polling serial port via ```mbedls``` does not work. As the port appears quite quickly after reflashing. But isn't ready for opening. Hence it gives error: 
```
[errno 16] Device or resource busy: '/dev/ttyACM0'
``` 
This PR retries opening serial port for ```polling_time``` times (supply by option ```--polling-time```).

Also it is observed that statement ```serial.Serial.break_condition = False``` raises ```[Errno 32] Broken pipe```. But it does not affect further use of ```serial.Serial``` object.

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
@@ -50,8 +50,8 @@ class SerialConnectorPrimitive(ConnectorPrimitive):
             self.port = serial_port
 
         startTime = time.time()
+        self.logger.prn_inf("serial(port=%s, baudrate=%d, timeout=%s)"% (self.port, self.baudrate, self.timeout))
         while time.time() - startTime < self.serial_pooling:
-            self.logger.prn_inf("serial(port=%s, baudrate=%d, timeout=%s)"% (self.port, self.baudrate, self.timeout))
             try:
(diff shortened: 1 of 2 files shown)


mazimkhan:
@adbridge @bridadan please review

mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
```diff
@@ -50,20 +49,26 @@ def __init__(self, name, port, baudrate, config):
...
-            self.LAST_ERROR = "connection lost, serial.Serial(%s, %d, %d): %s"% (self.port,
-                self.baudrate,
-                self.timeout,
-                str(e))
-            self.logger.prn_err(str(e))
-        else:
-            self.reset_dev_via_serial(delay=self.forced_reset_timeout)
+        startTime = time.time()
+        while time.time() - startTime < self.serial_pooling:
+            self.logger.prn_inf("serial(port=%s, baudrate=%d, timeout=%s)"% (self.port, self.baudrate, self.timeout))
```
adbridge: Assuming this is a print , do we really want to do this every time around the while ?

adbridge:
Other than the one comment left, the rest LGTM .

mbed_host_tests/host_tests_conn_proxy/conn_primitive_serial.py
mazimkhan: good point

PR #139, opened 2016-12-05
Log comparision feature

mazimkhan:
This feature adds option ```--compare-log``` to compare target serial with a templated comparision log. Please see ```README.md``` changes for details.

This is PR is open for discussion and adding more scenarios in consideration.

@adbridge @bridadan Please review.


mazimkhan:
Still need to check if we can change binary execution timeout. As different examples will take different execution time.

mazimkhan:
@bridadan ping

bridadan:
I read it about 5 times before I figured out exactly what this was doing and what it was trying to accomplish :smile: 

I this is pretty good! Easy to understand and pretty straight forward! Definitely a good start, maybe even good enough for all the examples. The thing I had in mind was more complicated then this, so I like the simplicity of this! I'll have to try it out a bit.

PR #137, opened 2016-11-22
Unification of poll_ing_timeout used in serial and remote conn primitive

mazimkhan:
Changes:
- Unify polling timeout use by serial and remote connection primitive
- Fix name from pooling to polling
- Fixes https://github.com/ARMmbed/htrun/issues/136
- Recommended change in greentea to pass the option ```--poll-time``` to htrun

PR #140, opened 2016-12-09
Fixes traceback caused by wrong variable reference

bridadan:
Looks like some references to `self.serial_pooling` were missed in #137. It's currently causing a traceback on master, this should fix it!


mazimkhan:
Thanks for fixing this.

PR #146, opened 2017-01-27
Handle regex exception when comparing log

mazimkhan:
Lines from template log are first matched literally and then as regex. In failure cases a line that is expected to match literally will not match and be tried as a regex patter. If the line has unqualified regex it may raise exception. Hence regex exception should be handled as a possibility and match() function should return False (no match) on exception.

@bridadan @adbridge Please review.

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -22,6 +22,7 @@ import re
 import sys
 import traceback
 from time import time
+from sre_compile import error
 from Queue import Empty as QueueEmpty   # Queue here refers to the module, not a class
 
 from mbed_host_tests import BaseHostTest
@@ -167,6 +168,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                 "sync_behavior" : self.options.sync_behavior,
                 "platform_name" : self.options.micro,
                 "image_path" : self.mbed.image_path,
+                "skip_reset": self.options.skip_reset,
             }
 
             if self.options.global_resource_mgr:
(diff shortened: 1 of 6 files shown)


adbridge:
@mazimkhan merge commits onto PRs are frowned upon (certainly for mbed-os) code as it makes patching hard. A rebase and force push would have been better....

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -527,6 +528,10 @@ def match_log(self, line):
         if self.compare_log_idx < len(self.compare_log):
             regex = self.compare_log[self.compare_log_idx]
             # Either the line is matched as is or it is checked as a regular expression.
-            if regex in line or re.search(regex, line):
-                self.compare_log_idx += 1
+            try:
+                if regex in line or re.search(regex, line):
+                    self.compare_log_idx += 1
+            except error:
```
bridadan: Any chance we can use the actual regex exception here? Is there one that you catch explicitly here? I don't think it's generally good practice to catch ALL exceptions and fail silently.

mazimkhan: Please see line https://github.com/ARMmbed/htrun/pull/146/files#diff-abaa341b8a118c20a730035041013df4R25
This is actually specific regex exception. I think the name "error" causes confusion.

mazimkhan:
@adbridge Done.

mbed_host_tests/host_tests_runner/host_test_default.py
bridadan: You're totally right, my bad! And I learned something new today :smile: But man yeah, what a blissfully ambiguous error name, good going regex library lol.

PR #150, opened 2017-04-03
cleanup remote dut initialization

jupe:
just cleanup a bit remote dut initialization phase

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
@@ -73,9 +73,8 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
 
         # Remote DUT connection, flashing and reset...
         try:
-            self.__remote_connect(baudrate=self.baudrate)
-            self.__remote_flashing(self.image_path)
             self.__remote_disconnect()
+            self.__remote_flashing(self.image_path)
             self.__remote_connect(baudrate=self.baudrate)
             self.__remote_reset()
         except Exception as e:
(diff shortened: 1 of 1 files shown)


mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -73,9 +73,8 @@ def __remote_init(self):
 
         # Remote DUT connection, flashing and reset...
         try:
-            self.__remote_connect(baudrate=self.baudrate)
```
jupe: this is just waste of time :)

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -73,9 +73,8 @@ def __remote_init(self):
...
-            self.__remote_flashing(self.image_path)
             self.__remote_disconnect()
```
jonikula: Does this disconnect need to be here, since we don't connect any more?

jupe: if check first if there was connection - if was then disconnect. this is neede to select right serialport configuration when connecting again..

jonikula: So the __remote_disconnect actually checks if there is a connection? If so, this is ok.

jonikula: Yea, it seems to have a check for this. lgtm, though I have no experience with htrun so I'll leave it for someone else to approve this.

PR #152, opened 2017-04-05
verify selected remote resource before using it

jupe:
check that remote resource exists before using it

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
@@ -86,6 +86,8 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
     def __remote_connect(self, baudrate=DEFAULT_BAUD_RATE, buffer_size=6):
         """! Open remote connection to DUT """
         self.logger.prn_inf("opening connection to platform at baudrate='%s, bufferSize=%d'"% (baudrate, buffer_size))
+        if not self.selected_resource:
+            raise Exception("remote resource not exists!")
         try:
             serial_parameters = self.remote_module.SerialParameters(lineMode=False, baudrate=baudrate, bufferSize=buffer_size)
             self.selected_resource.openConnection(parameters=serial_parameters)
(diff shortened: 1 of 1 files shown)


mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -144,6 +154,7 @@ def finish(self):
                     self.selected_resource.closeConnection()
                 if self.selected_resource.is_allocated:
                     self.selected_resource.release()
+                    self.selected_resource = None
```
bridadan: I think you should move this out of the `if` statement

jupe: reason I did like this was that if close connection or release throw exception there is still one chance to try - better to try again than forget because could be that if close connection fails in some reason it doesnt release at all and it might take time when it is finally released..

mazimkhan: Do you mean user will try again? Can we not make it part of this method. Like line https://github.com/ARMmbed/htrun/pull/152/files#diff-ab29c0280949087d36c3c6cca0af4441R151 can be 
```while self.selected_resource:```

bridadan: I like the retry idea!

jupe: > Do you mean user will try again?

actually not. This whole function is called twice by htrun - once from core, and once when this class destructor is called. So we actually retry already twice. Anyway this shouldn't normally fail ever. If we really want to do retry for this we should do retry for other calls as well and perhaps more intelligent logic for it -> I wouldn't like to add any retry logic here for now. Resource is released anyway after given expiration time in remote side so it doesn't cause big harm even this fails here.

bridadan: @jupe The whole "this function is called twice" I think is just for completeness, not design. It's called in the destructor in the off chance you haven't cleaned up the connection cleanly.

Sounds like this statement belongs where it is 👍

jupe:
ready to merge ?

bridadan:
Seems ok to me

PR #157, opened 2017-05-24
And result with sync results

theotherjimmy:
Should avoid dropping copy failures


mazimkhan:
Has this been tested in a staging setup?

theotherjimmy:
I ran it on my linux box. I'm not sure what you mean by staging setup.

mazimkhan:
I meant this running in local CI and working stably. But since it is Linux specific I don't think it affects Cambridge and Austin CIs. So I will merge it.

PR #158, opened 2017-05-26
Added sync failure status: mbed greentea module should be synced with…

deepikabhavnani:
@bridadan @mazimkhan @studavekar : Please review code

Many boards in CI fail just after sync, suggestion to which was re-flash firmware and re-try. Capturing SYNC failure as separate for such cases.

Note: Code changes in greentea should be pulled along with this request

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -119,6 +119,11 @@ def conn_process(event_queue, dut_event_queue, config):
         connector.finish()
         event_queue.put(('__notify_conn_lost', error_msg, time()))
 
+    def __notify_sync_failed():
+        error_msg = connector.error()
+        connector.finish()
+        event_queue.put(('__notify_sync_failed', error_msg, time()))
+
     logger = HtrunLogger('CONN')
     logger.prn_inf("starting connection process...")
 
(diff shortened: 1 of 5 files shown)


mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -130,6 +135,7 @@ def __notify_conn_lost():
     sync_behavior = int(config.get('sync_behavior', 1))
     sync_timeout = config.get('sync_timeout', 1.0)
     conn_resource = config.get('conn_resource', 'serial')
+    last_sync = 0
```
bridadan: Nit pick: Could you use `True` and `False` vs `0` and `1`? This is more consistent with the rest of the codebase.

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -270,8 +276,16 @@ def __send_sync(timeout=None):
                         sync_uuid_list.append(sync_uuid)
                         sync_behavior -= 1
                         loop_timer = time()
+                        #Sync behavior will be zero and if last sync fails we should report connection
+                        #lost
+                        if sync_behavior == 0:
+                            last_sync = 1
```
bridadan: `True` here

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -270,8 +276,16 @@ def __send_sync(timeout=None):
...
                     else:
                         __notify_conn_lost()
                         break
+            elif last_sync == 1:
```
bridadan: `True` here

adbridge:
LGTM

PR #159, opened 2017-05-31
Re-try flashing in case of flash failures.

deepikabhavnani:
Code change picked upfrom 'Brain Daniels' repo: copy_and_remount_refactor

@studavekar @mazimkhan @bridadan : Please review

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -500,7 +500,6 @@ class DefaultTestSelector(DefaultTestSelectorBase):
             if self.options.skip_flashing:
                 self.logger.prn_inf("copy image onto target... SKIPPED!")
             else:
-                result = None
                 self.logger.prn_inf("copy image onto target...")
                 result = self.mbed.copy_image()
                 if not result:
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -501,12 +501,8 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                 self.logger.prn_inf("copy image onto target... SKIPPED!")
             else:
                 result = None
-                for cnt in range(1, 4):
-                    self.logger.prn_inf("copy image onto target (try %s out of 3)..." % str(cnt))
-                    result = self.mbed.copy_image()
-                    if result:
-                        break
-
+                self.logger.prn_inf("copy image onto target...")
+                result = self.mbed.copy_image()
                 if not result:
                     result = self.RESULT_IOERR_COPY
                     return self.get_test_result_int(result)
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -500,8 +500,13 @@ class DefaultTestSelector(DefaultTestSelectorBase):
             if self.options.skip_flashing:
                 self.logger.prn_inf("copy image onto target... SKIPPED!")
             else:
-                self.logger.prn_inf("copy image onto target...")
-                result = self.mbed.copy_image()
+                result = None
+                for cnt in range(1, 4):
+                    self.logger.prn_inf("copy image onto target (try %s out of 3)..." % str(cnt))
+                    result = self.mbed.copy_image()
+                    if result:
+                        break
+
                 if not result:
                     result = self.RESULT_IOERR_COPY
                     return self.get_test_result_int(result)
--- a/mbed_host_tests/host_tests_runner/mbed_base.py
+++ b/mbed_host_tests/host_tests_runner/mbed_base.py
@@ -85,12 +85,14 @@ class Mbed:
                 self.logger.prn_err("Test configuration JSON Unexpected error:", str(e))
                 raise
 
-    def copy_image(self, image_path=None, disk=None, copy_method=None, port=None, retry_copy=3):
+    def copy_image(self, image_path=None, disk=None, copy_method=None, port=None, retry_copy=5):
         """! Closure for copy_image_raw() method.
         @return Returns result from copy plugin
         """
-
         def get_remount_count(disk_path, tries=2):
+            """! Get the remount count from 'DETAILS.TXT' file
+            @return Returns count, None if not-available
+            """
             for cur_try in range(1, tries + 1):
                 try:
                     files_on_disk = [x.upper() for x in os.listdir(disk_path)]
@@ -99,44 +101,29 @@ class Mbed:
                             for line in details_txt.readlines():
                                 if 'Remount count:' in line:
                                     return int(line.replace('Remount count: ', ''))
+                            # Remount count not found in file
+                            return None
+                    # 'DETAILS.TXT file not found
+                    else:
+                        return None
+
                 except OSError as e:
                     self.logger.prn_err("Failed to get remount count due to OSError.", str(e))
                     self.logger.prn_inf("Retrying in 1 second (try %s of %s)" % (cur_try, tries))
                     sleep(1)
+            # Failed to get remount count
+            return None
 
-        # Set-up closure environment
-        if not image_path:
-            image_path = self.image_path
-        if not disk:
-            disk = self.disk
-        if not copy_method:
-            copy_method = self.copy_method
-        if not port:
-            port = self.port
-        if not retry_copy:
-            retry_copy = self.retry_copy
-
-        flash_done = False
-        target_id = self.target_id
-
-        for count in range(0, retry_copy):
-            if flash_done:
-                break
-            initial_remount_count = get_remount_count(disk)
-
-            # Call proper copy method
-            result = self.copy_image_raw(image_path, disk, copy_method, port)
-            sleep(self.program_cycle_s)
-            if not result:
-                continue
-
+        def check_flash_error(target_id, disk, initial_remount_count):
+            """! Check for flash errors
+            @return Returns false if FAIL.TXT present, else true
+            """
             if not target_id:
                 self.logger.prn_wrn("Target ID not found: Skipping flash check and retry")
-                break
+                return True
 
             bad_files = Set(['FAIL.TXT'])
-
-                # Re-try at max 5 times with 0.5 sec in delay
+            # Re-try at max 5 times with 0.5 sec in delay
             for i in range(5):
                 # mbed_lstools.create() should be done inside the loop. Otherwise it will loop on same data.
                 mbeds = mbed_lstools.create()
@@ -168,14 +155,35 @@ class Mbed:
                             except IOError as error:
                                 self.logger.prn_err("Error opening '%s': %s" % (full_path, error))
 
-                            self.logger.prn_err("Error file contents:\n%s"% (bad_file_contents))
+                            self.logger.prn_err("Error file contents:\n%s" % bad_file_contents)
                         if common_items:
-                            result = False
-                        else:
-                            flash_done = True
-                        break
-                    sleep(0.5)
+                            return False
+                sleep(0.5)
+            return True
 
+        # Set-up closure environment
+        if not image_path:
+            image_path = self.image_path
+        if not disk:
+            disk = self.disk
+        if not copy_method:
+            copy_method = self.copy_method
+        if not port:
+            port = self.port
+        if not retry_copy:
+            retry_copy = self.retry_copy
+        target_id = self.target_id
+
+        for count in range(0, retry_copy):
+            initial_remount_count = get_remount_count(disk)
+            # Call proper copy method
+            result = self.copy_image_raw(image_path, disk, copy_method, port)
+            sleep(self.program_cycle_s)
+            if not result:
+                continue
+            result = check_flash_error(target_id, disk, initial_remount_count)
+            if result:
+                break
         return result
 
     def copy_image_raw(self, image_path=None, disk=None, copy_method=None, port=None):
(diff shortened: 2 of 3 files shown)


jupe:
why it can fail in first time? we have seen several flash burned out issues but reflashin doesn't help there..

bridadan:
@jupe It can fail for a number of reasons, not just flash failure. If the OS copies the binary file out of order, DAPLink will fail with an error. At this point it should be retried.

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -500,8 +500,13 @@ def execute(self):
             if self.options.skip_flashing:
                 self.logger.prn_inf("copy image onto target... SKIPPED!")
             else:
-                self.logger.prn_inf("copy image onto target...")
-                result = self.mbed.copy_image()
+                result = None
+                for cnt in range(1, 4):
```
bridadan: I would recommend breaking this out as a CLI parameter, not just hard coding 3 tries.

I realize that was my bad code so I guess this is cleaning time 😄

RomanSaveljev:
> It can fail for a number of reasons

@bridadan Could you elaborate more please? I do not think failure copying to MSD (be it provided by daplink or usb stick) is a normal case?

jupe:
> If the OS copies the binary file out of order

what does this mean?

what is other possible failure reasons?
If reflashing is needed it sounds a bit that it might actually
hide the root cause for failure which should be fixed rather. 
I would like to understand more about those corner cases.

mbed_host_tests/host_tests_runner/host_test_default.py
studavekar: i agree with @bridadan  on different failures, should we re-try only if there is a flashing failure like fail.txt, assert.txt is generated, We can log the failure as well so we don't miss what was the reason for failure.

We get fail.txt for all mbed supported devices, adding @c1728p9 .

bridadan:
@jupe @RomanSaveljev I understand your concern, but this is not to hide a deeper failure.

From past discussions with @c1728p9, DAPLink requires that all Mass Storage operations from the host OS to happen "in order". Meaning, when the OS sends the actual USB traffic representing the binary file, the OS sends it in chunks. Normally, the OS sends these in order, starting with the first chunk (ie. "data that should go in memory address 0") and ending with the last chunk (ie "the last byte in this chunk will be the last byte written to memory"). However, per the Mass Storage specification, the OS is allowed to send the data out of order if necessary. The OS will do this occasionally. I don't know the specific reasons for this, but @c1728p9 might know.

DAPLink requires the file to be sent in order because it doesn't have enough RAM to buffer the entire binary file before writing it to the target's flash. So if the OS sends the file out of order, the correct response to this is to just retry the flashing. Since we don't have any control over how the OS copies files, retrying is really the only option when flashing via the MSD. If we want to avoid this issue, we would need to flash over the debug channel.

@c1728p9 If I got anything wrong could you please correct me? Sorry for the wall of text, it's a fairly involved case 😄

bridadan:
>I do not think failure copying to MSD (be it provided by daplink or usb stick) is a normal case?

True, not by usb stick. And also usually not on DAPLink. However, it does happen, and we should try to catch these cases in automation.

c1728p9:
That all looks correct @bridadan. I have investigate these failures so if you need further details on what causes them (typically a disk cache flush) and how they can be recognized, let me know. A few common error messages in FAIL.TXT that can occur due to behavior out of DAPLink's control are:

* "An error occurred during the transfer" - Staring sector of the file changed during the transfer
* "The transfer timed out." - The complete file was never sent
* "File sent out of order by PC. Target might not be programmed correctly."

RomanSaveljev:
Thanks, the RAM limitation makes sense and this is valuable information. However, it raises a question to me whether MSD flashing is a proper choice for automation farm use case. No need to debate it right now, but I think we should look more into flashing with debugger.

bridadan:
**cough** pyOCD **cough** 😄

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -500,6 +500,7 @@ def execute(self):
             if self.options.skip_flashing:
                 self.logger.prn_inf("copy image onto target... SKIPPED!")
             else:
+                result = None
```
bridadan: Don't think this line is necessary since it will be replaced below by the call to `self.mbed.copy_image()`, any chance we can remove it?

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -92,10 +112,72 @@ def copy_image(self, image_path=None, disk=None, copy_method=None, port=None):
...
+                                print "FS_ERROR: Contents\n%s"% (bad_file_contents)
+                                if common_item != 'FAIL.TXT':
+                                    try:
+                                        os.remove(full_path)
+                                    except OSError as error:
+                                        print "ERROR removing '%s': %s" % (full_path, error)
+                            if common_items:
+                                result = False
+                                if 'ASSERT.TXT' in common_items:
+                                    raise Exception('ASSERT.TXT found!')
```
bridadan: Also, I forgot about these two lines: 

```
if 'ASSERT.TXT' in common_items:
    raise Exception('ASSERT.TXT found!')
```

These lines were originally added because we were debugging DAPLink issues. We shouldn't actually be raising an exception anymore, can you please remove these lines? (lines 174 and 175).

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -92,10 +112,72 @@ def copy_image(self, image_path=None, disk=None, copy_method=None, port=None):
             copy_method = self.copy_method
         if not port:
             port = self.port
+        if not retry_copy:
+            retry_copy = self.retry_copy
+
+        flash_done = False
+        target_id = self.target_id
+
+        for count in range(0,retry_copy):
```
bridadan: Nit pick: can you add a space between the `retry_copy` and `,`?

bridadan:
@mazimkhan This should be ready for your review now

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -92,10 +113,69 @@ def copy_image(self, image_path=None, disk=None, copy_method=None, port=None):
...
+                            except IOError as error:
+                                self.logger.prn_err("Error opening '%s': %s" % (full_path, error))
+
+                            self.logger.prn_err("Error file contents:\n%s"% (bad_file_contents))
+                        if common_items:
+                            result = False
+                        else:
+                            flash_done = True
+                        break
+                    sleep(0.5)
```
mazimkhan: This sleep should be at the level of ```if target_id in mbeds_by_tid:```. Because if mbedls does not detect this target then we quickly run out of retries.

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -92,10 +113,69 @@ def copy_image(self, image_path=None, disk=None, copy_method=None, port=None):
...
+                continue
+
+            if not target_id:
+                self.logger.prn_wrn("Target ID not found: Skipping flash check and retry")
+                break
+
+            bad_files = Set(['FAIL.TXT'])
+
+                # Re-try at max 5 times with 0.5 sec in delay
+            for i in range(5):
```
mazimkhan: Comments above above this line is not aligned with the indentation level.

mazimkhan: Also it would good to put this error checking in a separate function.

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -92,10 +113,69 @@ def copy_image(self, image_path=None, disk=None, copy_method=None, port=None):
...
+                # mbed_lstools.create() should be done inside the loop. Otherwise it will loop on same data.
+                mbeds = mbed_lstools.create()
+                mbeds_by_tid = mbeds.list_mbeds_by_targetid()   # key: target_id, value mbedls_dict()
+
+                if target_id in mbeds_by_tid:
+                    if 'mount_point' in mbeds_by_tid[target_id] and mbeds_by_tid[target_id]['mount_point']:
```
mazimkhan: After checking ```'mount_point' in mbeds_by_tid[target_id]``` it is redundant to check ```and mbeds_by_tid[target_id]```.

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -70,19 +75,35 @@ def __init__(self, options):
...
+                    files_on_disk = [x.upper() for x in os.listdir(disk_path)]
+                    if 'DETAILS.TXT' in files_on_disk:
+                        with open(os.path.join(disk_path, 'DETAILS.TXT'), 'r') as details_txt:
+                            for line in details_txt.readlines():
+                                if 'Remount count:' in line:
+                                    return int(line.replace('Remount count: ', ''))
+                except OSError as e:
+                    self.logger.prn_err("Failed to get remount count due to OSError.", str(e))
+                    self.logger.prn_inf("Retrying in 1 second (try %s of %s)" % (cur_try, tries))
+                    sleep(1)
```
mazimkhan: If it is required to retry when ```if 'DETAILS.TXT' in files_on_disk:``` is False, ```for line in details_txt.readlines():``` is empty and ```if 'Remount count:' in line:``` is false. Then we should have ```sleep``` in loop instead of exception block.
Else, for all above check we should return 0 or None.

mazimkhan: Missing return statement at the end of function or end of loop.

mazimkhan:
Comments added.

mbed_host_tests/host_tests_runner/mbed_base.py
bridadan: Can `mount_point` ever be in the mbed-ls listing but not valid? For instance, the board is present on the system but it is not mounted? (most likely a linux-specific scenario).

mazimkhan: Need to check. But here ```if mbeds_by_tid[target_id]``` check means if directory is not empty. If we pass ```if 'mount_point' in mbeds_by_tid[target_id]``` then ```if mbeds_by_tid[target_id]``` is always true.

bridadan: >If we pass `if 'mount_point' in mbeds_by_tid[target_id]` then `if mbeds_by_tid[target_id]` is always true.

That is true, but that isn't what the code is checking. The code is checking `if mbeds_by_tid[target_id]['mount_point']`, not `if mbeds_by_tid[target_id]`

mbed_host_tests/host_tests_runner/mbed_base.py
bridadan: @mazimkhan, after some offline discussion with @deepikabhavnani, we determined that correct behavior should be that if no `DETAILS.TXT` is detected OR if `DETAILS.TXT` is present but the `Remount count` line isn't present, then the function should return immediately. This is because both conditions indicate the firmware present on the board doesn't support the "Remount count" feature. If it isn't possible to determine these conditions due to an `OSError`, then we sleep for a bit to give the OS time to properly enumerate the device before trying again.

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -70,19 +75,92 @@ def __init__(self, options):
...
+                            bad_file_contents = "[failed to read bad file]"
+                            try:
+                                with open(full_path, "r") as bad_file:
+                                    bad_file_contents = bad_file.read()
+                            except IOError as error:
+                                self.logger.prn_err("Error opening '%s': %s" % (full_path, error))
+
+                            self.logger.prn_err("Error file contents:\n%s" % bad_file_contents)
+                        if common_items:
+                            return False
```
bridadan: Currently it seems like if nothing wrong is found with the current flashing operation (ex. remount count incremented and `FAIL.TXT` isn't found), then it will retry 5 times. Does this sounds right @deepikabhavnani?

If that's the case, then I think we need to make use of a flag here. Specifically, the "success" flag should be set in the `else` block of this `if` statement ("success" flag should default to `False`).

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -70,19 +75,92 @@ def __init__(self, options):
...
+                sleep(0.5)
+            return True
```
bridadan: In reference to my above comment, this function would then return the "success" flag here. This is important because it is possible that the loop will go through each of its iterations without making it to the end (for instance, by hitting the `continue` statements above).

mbed_host_tests/host_tests_runner/mbed_base.py
mazimkhan: Ok so we need to return if ```DETAILS.TXT``` is not present. We also need to specify what to return.

mbed_host_tests/host_tests_runner/mbed_base.py
mazimkhan: Ah sorry I missed the wrapped up line. Sorry:)

mbed_host_tests/host_tests_runner/mbed_base.py
deepikabhavnani: If FAIL.TXT is not found and everything seems fine than "check_flash_error" function will return True (success).

deepikabhavnani: '''
result = check_flash_error(target_id, disk, initial_remount_count)
 +            if result:
 +                break
'''
And we break in the return of function. We don't need flag and can use return of new function

PR #161, opened 2017-06-13
Fixed the return value of check_mount_point_ready

studavekar:
## Description
Currently the method check_mount_point_ready return false even if the mount point is ready.

## Status
**READY**

## Tests
- Have tests locally with mbed test and standlone mbedhtrun

@bridadan

### DIFF
--- a/mbed_host_tests/host_tests_plugins/host_test_plugins.py
+++ b/mbed_host_tests/host_tests_plugins/host_test_plugins.py
@@ -151,7 +151,7 @@ class HostTestPluginBase:
                 self.print_plugin_info("Mount point for '%s' changed from '%s' to '%s'..."% (target_id, destination_disk, new_destination_disk))
                 destination_disk = new_destination_disk
 
-        result = False
+        result = True
         # Check if mount point we've promoted to be valid one (by optional target_id check above)
         # Let's wait for 30 * loop_delay + init_delay max
         if not access(destination_disk, F_OK):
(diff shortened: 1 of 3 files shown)


mbed_host_tests/host_tests_plugins/host_test_plugins.py
```diff
@@ -151,7 +151,7 @@ def check_mount_point_ready(self, destination_disk, init_delay=0.2, loop_delay=0
                 self.print_plugin_info("Mount point for '%s' changed from '%s' to '%s'..."% (target_id, destination_disk, new_destination_disk))
                 destination_disk = new_destination_disk
 
-        result = False
+        result = True
         # Check if mount point we've promoted to be valid one (by optional target_id check above)
         # Let's wait for 30 * loop_delay + init_delay max
         if not access(destination_disk, F_OK):
```
mazimkhan: Shouldn't we just remove this if check and run the loop. If mount point is ready we just come out of the loop.

bridadan: Sounds ok to simplify it that way

studavekar: Sure simplifies logic can take out the condition, the only thing is will always print 
```
 "Waiting for mount point XX to be ready..

mazimkhan: You can move the print at https://github.com/ARMmbed/htrun/pull/161/files#diff-56032fd4ed339b468006a8a9ca289aa8R164

bridadan: I guess the whole point of that `if` statement is to prevent the printing of "Waiting for mount point..." if its not necessary.  Do you still want to make this change @mazimkhan?

mazimkhan: No problem you can leave it as is.

PR #163, opened 2017-08-07
Add option to request allocation based on specified tags

kegilbert:
In the process of integrating additional tests to the CI we realized it would be helpful to be able to tag particular devices in RaaS that had unique or additional setups for tests that required them (for example a K64F with an ESP8266 or Ethernet). 

## Usage 
Take the following tag as an example. 

image

There is a separate PR for changes to greentea to pipe the new argument to this level. Given that greentea change and the above tag setup, we could run tests with the following setup.

## Test Setup
```
mbedgt -g "K64F:raas_client:austin-ci-linux-001:8000" --test-spec=BUILD/tests/K64F/GCC_ARM/test_spec.json --report-html=K64F_GCC_ARM.html -e tests/host_tests/timing_drift_auto.py -v -V -n features-feature_lwip-tests-mbedmicro-net-tcp_hello_world --allocate-on-tag Ethernet                                                             
 ```

## Multiple Tags
Comma separated lists can also be supplied if multiple tags are required (device resource must match *all* of the supplied tags, it cannot be used as an 'or' list).
```
.... --allocate-on-tag Ethernet,ESP8266,...
```

### CODE REVIEW

**Comment by kegilbert** [comment]
cc @studavekar

File: mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py [hunk 1]
```diff
@@ -64,9 +69,15 @@ def __remote_init(self):
         # Query for available resource
         # Automatic selection and allocation of a resource
         try:
-            self.selected_resource = self.client.allocate({
-                "platform_name": self.platform_name
-            })
+            if self.tags:
```
jupe: if there is somethibg like `self.allocate_requirements` dict then we doesnt need these if-else statements... would be bit more simple and easier to further develop requirement parameters..

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -32,6 +31,12 @@ def __init__(self, name, config):
         self.baudrate = config.get('baudrate', DEFAULT_BAUD_RATE)
         self.image_path = config.get('image_path', None)
         self.polling_timeout = int(config.get('polling_timeout', 60))
+        self.tags = config.get('tags', None)
+
+        if(self.tags):
+            self.tags = {}
+            for tag in config.get('tags', None).split(','):
+                self.tags[tag] = True
```
jupe: tags has actually three states, True, False and None.. I propose to support at least both boolean types..

kegilbert: I left this as True for simplicity as it should be the only valid case for this situation (please correct me if I'm wrong).

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
kegilbert: Good point, let me know if the new implementation is an improvement.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
jupe: is `self.tags` needed somewhere? could this loop just add tags directly to requirements?

kegilbert:
@jupe Updated, I'll squash when the review is finished.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -31,14 +31,12 @@ def __init__(self, name, config):
         self.baudrate = config.get('baudrate', DEFAULT_BAUD_RATE)
         self.image_path = config.get('image_path', None)
         self.polling_timeout = int(config.get('polling_timeout', 60))
-        self.tags = config.get('tags', None)
         self.allocate_requirements = { "platform_name": self.platform_name }
 
-        if(self.tags):
-            self.tags = {}
+        if(config.get('tags', None)):
```
jupe: this looks much more clear :)

jupe:
By the way - this approach will use same tags as requirements for all tests that is going to be executed. Is that the way you really want ? It's going to be a bit inefficient (from utilization point of view)  in case that requirement is not really needed for all cases. Eg if we are going to run 100 tests, but only first test require tag X we are still using same device for rest of the tests even there could be "more cheaper" device in the pool.. This will be problematic in case that we have very little amount of devices with tag X than others and it is needed also for other test jobs...

kegilbert:
Sorry for the delay, the current plan is to generate a list of which tests require what tags based on the test binary dependencies and pass that along to htrun from the Greentea level. This would tune each greentea run to filter out device tags on a per-test/device level rather than passing it to an entire block of tests at the htrun layer.

theotherjimmy:
@kegilbert Is the aforementioned plan implemented in this PR? Is this ready for review?

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -32,6 +31,12 @@ def __init__(self, name, config):
         self.baudrate = config.get('baudrate', DEFAULT_BAUD_RATE)
         self.image_path = config.get('image_path', None)
         self.polling_timeout = int(config.get('polling_timeout', 60))
+        self.allocate_requirements = { "platform_name": self.platform_name }
+
+        if(config.get('tags', None)):
+            self.allocate_requirements["tags"] = {}
+            for tag in config.get('tags', None).split(','):
```
theotherjimmy: This `config.get` is not needed, as the above check ensures that the `tags` member is part of the config obejcet. I think it would be cleaner to just do:
```python
for tag in config.tags.split(','):
```

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -32,6 +31,12 @@ def __init__(self, name, config):
         self.baudrate = config.get('baudrate', DEFAULT_BAUD_RATE)
         self.image_path = config.get('image_path', None)
         self.polling_timeout = int(config.get('polling_timeout', 60))
+        self.allocate_requirements = { "platform_name": self.platform_name }
+
+        if(config.get('tags', None)):
```
theotherjimmy: Could you drop the parenthesis here? it's more idiomatic in python to omit parenthesis for simple checks.

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, options):
         self.serial_baud = DEFAULT_BAUD_RATE
         self.serial_timeout = 1
 
+        self.tags = self.options.allocate_on_tag
```
theotherjimmy: Is this assignment really needed? why not just use `self.options.allocate_on_tag` directly in place of `self.tags`?  Further, you should ensure that if an object is constructed, it has a valid `tags` attribute, so that you don't have to do odd `getattr` shenanigans later.

kegilbert: The other components of the config structure from [here](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_runner/host_test_default.py#L162) are defined above, the options structure that contains the tags at this level are not passed through the config structure [here](https://github.com/ARMmbed/htrun/pull/163/files#diff-ab29c0280949087d36c3c6cca0af4441R23).

Was that what you were referring to? I was not able to find another way to pass the existing options.allocate_on_tag data to the underlying RemoteConnectorPrimitive constructor.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -19,7 +19,6 @@
 from mbed_host_tests import DEFAULT_BAUD_RATE
 from mbed_host_tests.host_tests_conn_proxy.conn_primitive import ConnectorPrimitive
 
-
```
theotherjimmy: Please leave this line in.

kegilbert:
@theotherjimmy: @studavekar and I are still tinkering with the tag list generation implementation, updated what I had up but we still need to update a few things.

mbed_host_tests/host_tests_runner/mbed_base.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, options):
         self.serial_baud = DEFAULT_BAUD_RATE
         self.serial_timeout = 1
 
+        self.tags = self.options.tag_filters
```
theotherjimmy: Please drop this assignment, and the subsequent line. You are not using `self.tags` anywhere.

mbed_host_tests/tags
```diff
@@ -0,0 +1,402 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
theotherjimmy: This looks like ctags output or something. Is this file needed?

mbed_host_tests/host_tests_runner/mbed_base.py
kegilbert: Do we need the other self.option assignments above this as well? 
`self.retry_copy = self.options.retry_copy`, etc are part of the same argument flow. I can run without this line since we define the value initially when we create the [config dictionary](https://github.com/kegilbert/htrun/blob/ffbcba2d104bea6d5167e020897b893fb37d3e87/mbed_host_tests/host_tests_runner/host_test_default.py#L159), should I clean up the other assignments or just leave this line out?

mbed_host_tests/tags
kegilbert: Definitely not, my bad

mbed_host_tests/host_tests_runner/mbed_base.py
theotherjimmy: > should I clean up the other assignments or just leave this line out?

Yes, but could you keep it in a separate commit. That way it's easier to see that it's an unrelated refactor.

theotherjimmy:
@kegilbert Squashed onto master.

PR #165, opened 2017-08-22
Refactor code to be compatible with python3

scartmell-arm:
Fixed issues with python3 compatibility, most were highlighted by the 2to3 conversion tool.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: python
 python:
   - "2.7"
-  - "3.4"
+  - "3.5"
+  - "3.6"
 
 install:
   - pip install coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
 
 install:
   - pip install coverage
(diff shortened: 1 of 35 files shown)


theotherjimmy:
Nicely done @scartmell-arm. Could you also update the travis.yml to build this with python 3 as well?

theotherjimmy:
I just tried this. Could you add the encoding specifier to the `setup.py` open calls?

theotherjimmy:
@scartmell-arm Thanks for the update. I think this should be good to go, but Circle does not seem to be running the CI 3 times, each with a different version of python. That's annoying.

scartmell-arm:
Hm. Circle CI doesn't seem to support the travis build matrix. It only seems to be doable through their [Circle CI Workflows ](https://circleci.com/docs/2.0/workflows/) or third party [circleci-matrix](https://github.com/michaelcontento/circleci-matrix)

theotherjimmy:
It looks like the plan was to enable travis at some point... @mazimkhan is this the case? Or should we convert the .travis.yml into a circle CI file?

mazimkhan:
Some how the repo was not active on travis. Activated it now. But it may only trigger with another push.

theotherjimmy:
LGTM.  Merging.

PR #166, opened 2017-09-18
Wait forever in case global resource manager

mazimkhan:
Currently connection sub-process is started and ```htrun``` waits for ```process_start_timeout``` secs. This is not correct when global resource manager is used. As resource allocation can take really long time.

@bridadan @studavekar Please review. Required for use with RAAS.

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -227,7 +227,11 @@ class DefaultTestSelector(DefaultTestSelectorBase):
 
         conn_process_started = False
         try:
-            (key, value, timestamp) = event_queue.get(timeout=self.options.process_start_timeout)
+            # Wait for the start event. Process start timeout does not apply in
+            # Global resource manager case as it may take a while for resource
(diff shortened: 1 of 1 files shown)


adbridge:
@theotherjimmy Azim says this one is urgent can you ensure it moves through swiftly

theotherjimmy:
@mazimkhan What's up with that merge commit?

theotherjimmy:
I re-triggered Travis to check for Python 3 compatibility.

mazimkhan:
Removed merge commit.

theotherjimmy:
@mazimkhan Thanks! I'll merge this when testing passes.

theotherjimmy:
Should we also make specifying the timeout and the use of a global resource manager mutually exclusive options?

theotherjimmy:
Testing went green. I'm going to merge this, and I'll investigate my own question above.

mazimkhan:
Can we release htrun please. So that we can get it in Oulu CI.

bridadan:
Y'all are too quick for me 😄

PR #168, opened 2017-10-05
add args --sync-timeout parameter delay between __sync packets

studavekar:
Currently this argument was used but never set, always defaulting to 1 sec. In
case of RaaS and when device boot takes longer we end up of sending multiple
sync_packet causing few test to fail like echo test.

### DIFF
--- a/mbed_host_tests/__init__.py
+++ b/mbed_host_tests/__init__.py
@@ -202,7 +202,7 @@ def init_host_test_cli_params():
                       dest="sync_timeout",
                       default=5,
                       type=int,
-                      help="Define delay between between __sync packet (Default is 5 seconds)",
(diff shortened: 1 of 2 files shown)


mbed_host_tests/__init__.py
```diff
@@ -198,6 +198,13 @@ def init_host_test_cli_params():
                       help="Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of times (Default 2 time)",
                       metavar="SYNC_BEHAVIOR")
 
+    parser.add_option("", "--sync-timeout",
+                      dest="sync_timeout",
+                      default=5,
+                      type=int,
+                      help="Define delay between between __sync packet (Default is 5 seconds)",
```
theotherjimmy: Could you add units here? "delay in seconds" or similar.

PR #174, opened 2017-11-16
Allow a host test to reset the device under test

c1728p9:
Add a host test reset function to allow the current DUT to be reset. This allow unexpected power loss to be tested.

CC @bridadan @studavekar @theotherjimmy

### DIFF
--- a/mbed_host_tests/host_tests/base_host_test.py
+++ b/mbed_host_tests/host_tests/base_host_test.py
@@ -64,6 +64,14 @@ class BaseHostTestAbstract(object):
         if self.__event_queue:
             self.__event_queue.put(('__reset_dut', value, time()))
 
+    def reset(self):
+        """
+        Reset the device under test and continue running the host test
+        :return:
+        """
+        if self.__event_queue:
+            self.__event_queue.put(("__reset", "0", time()))
+
     def notify_conn_lost(self, text):
         """! Notify main even loop that there was a DUT-host test connection error
         @param consume If True htrun will process (consume) all remaining events
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -231,10 +231,15 @@ def conn_process(event_queue, dut_event_queue, config):
         else:
             # Return if state machine in host_test_default has finished to end process
             if key == '__host_test_finished' and value == True:
-                logger.prn_inf("received special even '%s' value='%s', finishing"% (key, value))
+                logger.prn_inf("received special event '%s' value='%s', finishing"% (key, value))
                 connector.finish()
                 return 0
-            if not connector.write_kv(key, value):
+            elif key == '__reset':
+                logger.prn_inf("received special event '%s', resetting dut" % (key))
+                connector.reset()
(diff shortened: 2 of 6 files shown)


mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -64,6 +64,14 @@ def reset_dut(self, value):
         if self.__event_queue:
             self.__event_queue.put(('__reset_dut', value, time()))
 
+    def reset(self):
```
bridadan: I know I wrote this code pretty hastily, but let's get the keys matching the function names if at all possible.

Can we make this function `soft_reset_dut`?

This would mean you'd need to update your custom host test as well.

mbed_host_tests/host_tests/base_host_test.py
```diff
@@ -64,6 +64,14 @@ def reset_dut(self, value):
...
+        """
+        Reset device under test
```
bridadan: Let's make this comment different from the function above. Something like

> Reset device under test (without restarting the host test)

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -231,10 +231,15 @@ def __send_sync(timeout=None):
         else:
             # Return if state machine in host_test_default has finished to end process
             if key == '__host_test_finished' and value == True:
-                logger.prn_inf("received special even '%s' value='%s', finishing"% (key, value))
+                logger.prn_inf("received special event '%s' value='%s', finishing"% (key, value))
                 connector.finish()
                 return 0
-            if not connector.write_kv(key, value):
+            elif key == '__reset_dut':
```
bridadan: Let's make this key match the event queue key: `__reset_dut` -> `__soft_reset_dut`.

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -366,6 +366,9 @@ def process_code_coverage(key, value, timestamp):
                         self.logger.prn_inf("%s(%s)" % (key, str(value)))
                         result = value
                         event_queue.put(('__exit_event_queue', 0, time()))
+                    elif key == '__soft_reset_dut':
+                        # This event only resets the dut, not the host test
+                        dut_event_queue.put(('__reset_dut', True, time()))
```
bridadan: Let's make this key match the event queue key: `__reset_dut` -> `__soft_reset_dut`.

c1728p9:
PR in little FS for this change:
https://github.com/ARMmbed/mbed-littlefs/pull/10

theotherjimmy:
@c1728p9 Thanks! Could you update the function names as @bridadan recommended? After that, a clean test run is all that's required.

theotherjimmy:
@c1728p9 Will we need to release a new version of htrun after this patch goes in?

bridadan:
Hi @theotherjimmy I forgot to post an update here but offline we decided to just go with the `reset` naming instead of the `soft_reset_dut` since it's not really a soft reset. There is an existing function called `reset_dut`, however I'm not currently aware of anything that's actually using it. And yes we'll need to get a new version out at some point.

theotherjimmy:
Thanks @bridadan. Sorry for the noise guys.

PR #178, opened 2017-12-01
Raas staging

theotherjimmy:


theotherjimmy:
Looks like all of this made it to master at some point

PR #181, opened 2018-01-15
Keep the same extension of the executable files copied to MPS2

hug-dev:
To allow to write .elf file path in the test_spec.json file, this patch
keeps the same extension when copying mbed.extension file onto the MPS2
board.

Signed-off-by: Hugues de Valon <hugues.devalon@arm.com>

### DIFF
--- a/mbed_host_tests/host_tests_plugins/module_copy_mps2.py
+++ b/mbed_host_tests/host_tests_plugins/module_copy_mps2.py
@@ -47,7 +47,9 @@ class HostTestPluginCopyMethod_MPS2(HostTestPluginBase):
         @return Returns True if copy (flashing) was successful
         """
         result = True
-        destination_path = os.path.join(destination_disk, "mbed.bin")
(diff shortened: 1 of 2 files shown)


ashok-rao:
@bridadan @theotherjimmy .. bump!

theotherjimmy:
That looks like someone made an assumption that turned out to be invalid. @ashok-rao Could you write a unit test for this? (You're welcome to say "no")

hug-dev:
I think that the name of the function and its comment say that the file will always be copied as `mbed.bin`, see below:
```C
     def mps2_bin_copy(self, image_path, destination_disk):
         """! mps2 copy method for "mbed enabled" devices.
            this copies the binary always as mbed.bin
```
I am going to update those as well to reflect the good behaviour and add a unit test.

theotherjimmy:
Thanks @hug-dev ! If you would like any help with that unit test please ask!

theotherjimmy:
Coveralls numbers look good. @bridadan Does the test look okay to you?

PR #186, opened 2018-03-20
Fix echo tests and set expiration timeout to 600 sec

studavekar:
These are stability fixes 

1) Fix echo tests
2) set expiration  timeout 600 secs

### DIFF
--- a/mbed_host_tests/host_tests/echo.py
+++ b/mbed_host_tests/host_tests/echo.py
@@ -18,6 +18,7 @@ limitations under the License.
 
 import uuid
 from mbed_host_tests import BaseHostTest
+import time
 
 class EchoTest(BaseHostTest):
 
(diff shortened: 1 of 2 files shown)


studavekar:
circleci failure doesn't look like related to this change 
```
  File "/usr/lib/python2.7/unittest/loader.py", line 100, in loadTestsFromName
    parent, obj = obj, getattr(obj, part)
AttributeError: 'module' object has no attribute 'mps2_reset'
```

jupe:
default expiration is one hour in server side, so this now will reduce it to 5 min which might causes unexpectedly failures for tests which takes longer.. If original reason for smaller expiration was ensure that resources get released sooner in corner cases maybe this is kind of reasonable but because even flashing could take more than 5 min in worst scenario default value should be much longer. I would stay in default value still.. Perhaps original problem could be avoided with #189..

theotherjimmy:
@studavekar Rebase/merge needed.

theotherjimmy:
@studavekar I'm closing this as I can't take it without that rebase/merge. Please reopen if you still need it.

PR #187, opened 2018-05-11
add "--fm" option  to support FastModels in mbedhtrun

jamesbeyond:
# Description

In order to support FastModels in mbed testing framework, an extra option "--fm" been added to mbedhtrun.

This option will allow mbedhtrun hooks up with fm_agent module, and run tests over FastModels

The command line format of the --fm is following the RaaS_Client "--grm" convention

# Details
* "--fm" been added to htrun
* user could use __mbedhtrun -m \<fastmodel> --fm <config_name> -f <test_image>__
  * e.g.  __mbedhtrun -m FVP_MSP2_M3 --fm DEFAULT -f ticker.elf__
* conn_primitive_fastmodel class been added, allow communication between htrun and fm_agent

This PR is related with other PRs regarding to FastModel support in testing framework 
ARMmbed/greentea#270
ARMmbed/mbed-ls#345

CC. @theotherjimmy @studavekar @bridadan @bulislaw


mbed_host_tests/__init__.py
```diff
@@ -282,6 +282,10 @@ def init_host_test_cli_params():
                       dest='global_resource_mgr',
                       help='[Experimental] Global resource manager service module name, IP and port, example remote_client:10.2.123.43:3334')
 
+    parser.add_option('-s', '--srm',
+                      dest='simulator_resource_mgr',
+                      help='[Experimental] Simulator resource manager service module name, and a config file, example fastmodel_agent:DEFAULT')
```
bridadan: Since the srm implements the same connector primitive as the grm, could we not merge the two arguments? Does it mostly come down to the fact that the srm doesn't require the IP address and port? If so, we could probably relax that requirement of the grm.

jamesbeyond: @bridadan I initially thought about using the same --grm option. But after working on it for a while, I realized it is not very practical.

Here  are some of my concerns:
1. The required parameters are different: --grm need a server IP and port number to the remote server. --srm just require a config_file.
2.  The internal logic of connection is different: --grm will need query the server for an available target. if the target is used by others, it will wait for it been released. --srm not having this logic at all. 
3. the target behavior is different: --grm still use the real hardware via serial, and need to flash and reset the target. --srm will need to launch the model first before connecting to it and use model-specific load/run function, also serial in replaced via socket. 

So I can image, even we can get them merged into the same primitive connector, most of the code will use lost of if/else redirect them to a totally different route.
And not mention that will require both code change from htrun and RaaS-pyclient, which will be risky for broken CI test. (As I know lots of CI using --grm interface, and the command hard-coded in the Jenkins side)

bridadan: Thanks for raising those points, I'll go through them a bit:

1. I think we could probably relax this requirement on the `--grm` argument and leave it to the module to define their own arguments. This wouldn't be a breaking change for the existing modules in use and it could open the door for the fastmodel_agent.

2. While the internal logic is different, I don't think that is all that important from the perspective of the user. You can supply anything as a "global resource manager", including simulators.

3. I'd say probably the same thing I said in `#2`. What is actually happening isn't really important for the use to know. If they are supplying the module on the command line, then I think we can expect that they are familiar with what is happening on the other end 😄 

And for your last point:
> And not mention that will require both code change from htrun and RaaS-pyclient, which will be risky for broken CI test

I don't think it'll require any changes from the raas-pyclient, just changes on the htrun side. And we can write some test cases ahead of time to be sure we don't break it. The interface is relatively simple so I think should be quite doable.

I realize this is all extending the integration time for your component here, is this time sensitive?

jamesbeyond: I see your point.
So do you mean change both the connection types share the same --grm command line option, but use different primitive connectors? I think that's possible, and fairly easy change.

Currently, the function for parsing --grm option is in htrun : https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_runner/host_test_default.py#L180-L188
If we want to 
> leave it to the module to define their own arguments.

we have to move such function into RaaS_clinet or fastmodel_agent module.
That why I said it will need code changes into RaaS_client.

Otherwise, we will need leave this in the htrun, which means determin the arguments depens on raas_client or fastmodel_agent

jamesbeyond: @bridadan for timing wise.
Actually I don't know how is the testing tools keep up with mbed-os release.
I would really like to get this merged before 5.9.

I didn't any tags on htrun.
which I guess it is going to be a bit more flexable compare to mbed-os, which will need to have the code freeze by the end of next week for 5.9?

bridadan: >we have to move such function into RaaS_clinet or fastmodel_agent module.
That why I said it will need code changes into RaaS_client.

Oh duh, you're right! Yeah so maybe that's a bit more out of scope than I thought. I'd like to hear from @theotherjimmy before we make any definite decisions on the path forward.

Yeah for releases, htrun will typically trail Mbed OS a little bit, so we can always release this at 5.9 or a little after.

theotherjimmy: The differing parameters are the thing that gets it for me. 2 methods of doing the same thing are bad if they take the same parameters, 2 implementation of an interface are good when they're constructed from differing params.

theotherjimmy: `--srm` is okay. Maybe we should consider a different switch name?

bridadan: I don't actually have a problem with the name. And I've thought a bit more about the current implementation and it really seems ok as is. I was kind of worried that we were duplicating stuff, but that's something we could always address in the future.

jamesbeyond: Hi guys @theotherjimmy @bridadan,
Is there any conclusion on this PR?

Do you think is the switch "-s, --srm" appropriate ?
current usage is *--srm <module_name>:<config_name>*
*e.g. --srm fm_agent:DEFAULT*
I started design it like this, becasue I thought other people could like write other simulator module for their own models.

Now I start to favor a new switch without module_name, module_name is hard coded inside htrun
*e.g. --fm <config_name>*

jamesbeyond: BTW. what's the error of circleci about?
It seems passed on my laptop.

bridadan: I'd be fine with moving to your proposed `--fm`. Makes it a little simpler!

For the circleci failure, I don't think that's relevant to your changes.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_fastmodel.py
```diff
@@ -0,0 +1,156 @@
...
+            self.__fastmodel_load(self.image_path)
+            self.__fastmodel_run()
+
+
+    def __fastmodel_init(self):
+        """! Initialize models using fm_agent APIs """
+        self.logger.prn_inf("Initializing FastModel...")
+
+        try:
+            self.fm_agent_module = __import__("fm_agent")
```
bridadan: If we're going to know the name of the python module and we intend to release the python module on pip, can we go ahead and just add this as a dependency? The python module itself doesn't have any heavy dependencies correct?

Do you have an opinion here @theotherjimmy?

theotherjimmy: My opinion is exactly that.

jamesbeyond: I wouldn't suggest put the `fm_agent` module as a dependency to htrun for now, because:
1. `fm_agent` module do have another dependecy module, which is closed source. So I don't think it will be published in pip
2. `fm_agent`  module has some manual configuration steps need to be done by the users before it can be functional.
3. Not every user would be interested in using fast models. some people would like just use greentea/htrun with their own hardware, so they don't bother about these thing. htrun can used as it was.

bridadan: 1. Fair point, though if `fm_agent` is included by default at least now they only have to worry about one "optional" module.
2. That's good to know, we may want to revist this. If the info is simple enough, we may be able to use an environment variable OR we can make it part of the greentea/htrun CLI.
3. Again, I wouldn't worry too much about this. If it isn't a heavy dependency and it doesn't break the existing interface then I don't think this is a problem. It can just be there and not do any harm 😄

jamesbeyond: Hi guys, @bridadan @theotherjimmy 
I am back from a week vacation. So is there a conclusion on this matter yet?

I think the fundamental question is, do we really want htrun tightly bound with fm_agent module?  htrun will not run without it (even for people have no interest in using fast models ).

Or do we want to decouple them? So htrun will work as it was, on normal targets. evern fm_agent is not present.

bridadan: Hi @jamesbeyond, thanks for following up. @theotherjimmy and I were just discussing this offline.

What we determined was that since `fm_agent` is dependent on the commercial `fm` module (not installable through the pypi registry) that comes with the FastModel product, then I think what we'll want to do is conditionally expose the `--fm` option in both htrun and greentea based on if the `fm` module is installed/importable.

Perhaps this functionality can be added to `fm_agent` and greentea and htrun can just call into that module to see if everything is installed? Not sure of the specifics here, I'd be curious to hear your thoughts as well as @theotherjimmy's.

jamesbeyond: Hi @bridadan, thanks for getting back so quick.

I think the current changes do exactly what you said: 
the fm_agent module checks those required depends of fast-models to be installed.
htrun just call to use the modules.

BTW, This 2 PRs have been there for some time, I am really like to get them merged before 5.9 close up. Do you think there is enough time for it? or maybe we should speed it up a little bit?

Or Is this PR will still require other stakeholders to review or more CIs to pass? this is my first PR to the tools, I am not sure about the procedures yet.

theotherjimmy: @jamesbeyond Not quite. I would like the `--fm` switch to only be present when the fm_agent module is installed. As this PR stands, the `--fm` switch is always available.

jamesbeyond: Haa! I see the point.
That's not a big change. I will get that in ASAP.

Any comments on the timing to merge?

theotherjimmy: 10 min after that patch!

bridadan: @jamesbeyond I think I was saying something different than @theotherjimmy, sorry for the confusion! We synced offline and I wanted to post those conclusions here just so we're sure we're on the same page:

1. Let's leave `fastmodel_agent` as an optional install, just like you do now. Since it requires the extra configuration step and the `fm` module, we'll leave that up to the user to take care of before using htrun/greentea.
2. We would like the `--fm` switch to only be enabled when `fastmodel_agent` is installed/importable. So the `parser.add_option('', '--fm', ...` call will need to be conditional. This is to reduce users exposure to this unless they have `fastmodel_agent` installed and configured.

jamesbeyond:
Hi @bridadan @theotherjimmy 
Both Htrun and Greentea got updated.
The "--fm" option only shows only if it detects `fm_agent` module installed.
Also the README got updated as well.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_fastmodel.py
jamesbeyond: Thanks to both of you for the clarification. didn't expect the merge is that quick ;-)

PR #189, opened 2018-05-26
Several improvements for Global Resource Manager connection

jupe:
## Status
**IN PROGRESS**

Relates to MBEDOSTEST-37

## Changes:
* remove obsolete API usages and use new one
* fixes bug from read in case of reading fails
* fixes finish to releasing resource in case of close_connection fails
* fix connected function
* connected() validates that resource is also allocated
* validates that write() success
* release resource if flash or connection fails
* catch generic Exception when allocate fails
* validates that resource is allocated and connected before writing/reading it
* validates that resource is allocated before releasing
* improve traces
* pylint cleaning
* add basic tests

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
@@ -21,7 +21,7 @@ from mbed_host_tests.host_tests_conn_proxy.conn_primitive import ConnectorPrimit
 
 
 class RemoteConnectorPrimitive(ConnectorPrimitive):
-    def __init__(self, name, config):
+    def __init__(self, name, config, importer = __import__):
         ConnectorPrimitive.__init__(self, name)
         self.config = config
         self.target_id = self.config.get('target_id', None)
@@ -44,14 +44,14 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
         self.client = None
 
         # Initialize remote resource manager
-        self.__remote_init()
+        self.__remote_init(importer)
 
-    def __remote_init(self):
+    def __remote_init(self, importer):
         """! Initialize DUT using GRM APIs """
 
         # We want to load global resource manager module by name from command line (switch --grm)
         try:
-            self.remote_module = __import__(self.grm_module)
+            self.remote_module = importer(self.grm_module)
         except ImportError as error:
             self.logger.prn_err("unable to load global resource manager '%s' module!" % self.grm_module)
             self.logger.prn_err(str(error))
@@ -95,7 +95,7 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
         try:
             serial_parameters = self.remote_module.SerialParameters(baudrate=baudrate)
             self.selected_resource.open_connection(parameters=serial_parameters)
-        except self.remote_module.resources.ResourceError as error:
+        except Exception as error:
             self.logger.prn_inf("open_connection() failed")
             raise error
 
@@ -105,7 +105,7 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
         try:
             if self.connected():
                 self.selected_resource.close_connection()
-        except self.remote_module.resources.ResourceError as error:
+        except Exception as error:
             self.logger.prn_err("RemoteConnectorPrimitive.disconnect() failed, reason: " + str(error))
 
     def __remote_reset(self):
@@ -131,7 +131,7 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
         data = str()
         try:
             data = self.selected_resource.read(count)
-        except self.remote_module.resources.ResourceError as error:
+        except Exception as error:
             self.logger.prn_err("RemoteConnectorPrimitive.read(%d): %s" % (count, str(error)))
         return data
 
@@ -143,7 +143,7 @@ class RemoteConnectorPrimitive(ConnectorPrimitive):
                 if log:
                     self.logger.prn_txd(payload)
                 return True
-            except self.remote_module.resources.ResourceError as error:
+            except Exception as error:
                 self.LAST_ERROR = "remote write error: %s" % str(error)
                 self.logger.prn_err(str(error))
         return False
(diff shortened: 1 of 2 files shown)


jupe:
ci/circleci failure doesn't seems to related to this PR

jupe:
There is probably still something I did not fixed/recognised so need good reviewing :)

jupe:
@OPpuolitaival do you want to try this out in e2e perspective :) ?

OPpuolitaival:
@jupe I cannot test now. @jupe do we have new RAAS available if @studavekar can do some testing for this?

jupe:
this does not related to any specific server version, only client side..

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -71,38 +72,41 @@ def __remote_init(self):
         # Automatic selection and allocation of a resource
         try:
             self.selected_resource = self.client.allocate(self.allocate_requirements)
-
-        except self.remote_module.resources.ResourceError as e:
-            self.logger.prn_err("can't allocate resource: '%s', reason: %s"% (self.platform_name, str(e)))
+        except Exception as error:
```
bridadan: Typically we don't want to do general "catch all" exceptions. Is there not a more specific exception you can "except" on?

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -71,38 +72,41 @@ def __remote_init(self):
...
 
     def __remote_disconnect(self):
         if not self.selected_resource:
             raise Exception("remote resource not exists!")
-        if self.selected_resource.is_connected:
-            self.selected_resource.closeConnection()
+        try:
+            if self.connected():
+                self.selected_resource.close_connection()
+        except Exception as error:
```
bridadan: Same concern here regarding `Exception`.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -114,51 +118,59 @@ def __remote_reset(self):
...
         if not self.selected_resource.flash(filename, forceflash=forceflash):
             raise Exception("remote resources flashing failed!")
 
     def read(self, count):
         """! Read 'count' bytes of data from DUT """
-        if not self.selected_resource:
+        if not self.connected():
             raise Exception("remote resource not exists!")
-        date = str()
+        data = str()
```
bridadan: Ew, good catch!

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -114,51 +118,59 @@ def __remote_reset(self):
...
-            if log:
-                self.logger.prn_txd(payload)
-        return True
+        if self.connected():
+            try:
+                self.selected_resource.write(payload)
+                if log:
+                    self.logger.prn_txd(payload)
+                return True
+            except Exception as error:
```
bridadan: `Exception` issue here as well.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -114,51 +118,59 @@ def __remote_reset(self):
...
-            self.selected_resource,
-            self.selected_resource.is_connected])
+        return self.allocated() and self.selected_resource.is_connected
+
+    def __remote_release(self):
+        try:
+            if self.allocated():
+                self.selected_resource.release()
+                self.selected_resource = None
+        except Exception as error:
```
bridadan: `Exception` issue here as well.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
jupe: I totally agree that generic catch should be avoided. When I did these changes I was thinking same thing and how to tackle case when exception class name is client specific. Maybe only way to avoid generic catch is to specify exception class name to genric and document it.. that would be breaking change..

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -71,38 +72,41 @@ def __remote_init(self):
...
+        self.logger.prn_inf("opening connection to platform at baudrate='%s'" % baudrate)
         if not self.selected_resource:
             raise Exception("remote resource not exists!")
         try:
-            serial_parameters = self.remote_module.SerialParameters(lineMode=False, baudrate=baudrate, bufferSize=buffer_size)
-            self.selected_resource.openConnection(parameters=serial_parameters)
-        except self.remote_module.resources.ResourceError as e:
-            self.logger.prn_inf("openConnection() failed")
-            raise e
+            serial_parameters = self.remote_module.SerialParameters(baudrate=baudrate)
```
studavekar: @jupe thanks for taking out obsolete parameters.

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
```diff
@@ -114,51 +118,59 @@ def __remote_reset(self):
...
-                if self.selected_resource.is_connected:
-                    self.selected_resource.closeConnection()
-                if self.selected_resource.is_allocated:
-                    self.selected_resource.release()
-                    self.selected_resource = None
-            except self.remote_module.resources.ResourceError as e:
-                self.logger.prn_err("RemoteConnectorPrimitive.finish() failed, reason: " + str(e))
+        if self.allocated():
+            self.__remote_disconnect()
+            self.__remote_release()
```
studavekar: 👍

mbed_host_tests/host_tests_conn_proxy/conn_primitive_remote.py
jupe: Actually I think current approach is just good in this case. All generic try-catch are very restricted so that they catch only client call exceptions. There is no any "wide" try-catches in this file. Also this way we doesn't need to document here specific exception class names - it's enough that client throws any exception which based on `Exception`. And bonus - doesn't need to change any client libraries..  ;) . @bridadan  what do you think ?

bridadan: I don't have a better idea so I suppose it's ok for now!

PR #191, opened 2018-06-08
Fixing incorrect inclusion of fastmodel as default connproxy

bridadan:
This should fix https://github.com/ARMmbed/greentea/issues/274#issuecomment-395660594.

Could you take a quick look @jamesbeyond?


theotherjimmy:
Oh, that's bad.

jamesbeyond:
How could I missed that! My mistake!

Issue #193, opened 2018-06-12
handle faulty SYNC packets 


studavekar:
### Description

When HTRUN encounters a sync packet from previous tests, It state machine goes in weird state. It starts the tests however still continues to send sync packets and eventully fails with `{{result;timeout}}` or `{{result;sync_failed}}`

It should readout/flash any stream of data in communication channel before starting the test.

```
00:47:55.318 	calling mbedhtrun: mbedhtrun -m NRF51_DK -p DUMMY:9600 -f "BUILD/tests/NRF51_DK/ARM/TESTS/mbedmicro-mbed/attributes/attributes.hex" -e "TESTS/host_tests" --grm raas_client:austin-ci-linux-001:8000 -P 120 -C 6 --sync 10
00:47:55.318 mbedgt: mbed-host-test-runner: started
00:47:55.561 [1528825492.12][HTST][INF] host test executor ver. 1.3.1
00:47:55.561 [1528825492.12][HTST][INF] copy image onto target... SKIPPED!
00:47:55.563 [1528825492.12][HTST][INF] starting host test process...
00:47:55.566 [1528825492.12][CONN][INF] starting connection process...
00:47:55.567 [1528825492.12][CONN][INF] notify event queue about extra 120 sec timeout for serial port pooling
00:47:55.567 [1528825492.12][CONN][INF] initializing global resource mgr listener...
00:47:55.567 [1528825492.12][HTST][INF] setting timeout to: 120 sec
00:47:55.601 [1528825492.16][GLRM][INF] remote resources initialization: remote(host=austin-ci-linux-001, port=8000)
00:47:55.602 [1528825492.16][urllib3.connectionpool]Starting new HTTP connection (1): austin-ci-linux-001
00:47:55.606 [1528825492.16][urllib3.connectionpool]http://austin-ci-linux-001:8000 "POST /login HTTP/1.1" 200 213
00:47:55.609 [1528825492.16][urllib3.connectionpool]Starting new HTTP connection (1): austin-ci-linux-001
[...]

PR #200, opened 2018-06-28
Reset device after bad sync

aashishc1988:
To solve issue: https://github.com/ARMmbed/htrun/issues/193

Solution: 

The host would send an initial sync, and if it receives a bad sync it will reset the part and wait until a certain duration  before sending the second sync packet. Meanwhile, if it receives another sync packet from DUT (if the buffer was not empty), it will reset the part again. 
Corner case: The DUT has 4 bad syncs in its buffer, and when it sent out the 3rd one, host received it and told it to reset, and then host itself reached a stage where it is ready to send second sync. At this point, device sends out the 4th sync and begins executing the tests (since there is no ack from host), and this test is one of those kind where it would continuously spew out data to host via the buffer, in this case, sending second sync would have no effect because that stage has already passed. At this point, host will not receive any echo back from device and it would eventually time out. To solve this particular corner case, while sending second sync, we should reset the device first and then send second sync packet.

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -135,6 +135,7 @@ def conn_process(event_queue, dut_event_queue, config):
 
     logger = HtrunLogger('CONN')
     logger.prn_inf("starting connection process...")
+    
 
     # Send connection process start event to host process
     # NOTE: Do not send any other Key-Value pairs before this!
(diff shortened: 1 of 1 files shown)


aashishc1988:
Yes, I think that is a good idea :) Every time a faulty sync comes, reset the timeout to send fresh sync and keep listening for another sync (if any) from client, and then eventually send the new sync.

aashishc1988:
@bridadan Added the fix.

PR #203, opened 2018-07-06
disable timeout during consume_preamble_events

alekla01:
This pr disables **timeout in host_test_default** during setup (allocation, flashing, etc..).
 * disable **timeout** during setup (consume_preamble_events=True)
 * enable **timeout** during testing (consume_preamble_events=False)

related: MBEDOSTEST-41

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -88,10 +88,6 @@ def conn_primitive_factory(conn_resource, config, event_queue, logger):
     @param logger Host Test logger instance
     @return Object of type <ConnectorPrimitive> or None if type of connection primitive unknown (conn_resource)
     """
-    polling_timeout = int(config.get('polling_timeout', 60))
-    logger.prn_inf("notify event queue about extra %d sec timeout for serial port pooling"%polling_timeout)
-    event_queue.put(('__timeout', polling_timeout, time()))
-
     if conn_resource == 'serial':
         # Standard serial port connection
         # Notify event queue we will wait additional time for serial port to be ready
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -263,7 +263,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
         try:
             consume_preamble_events = True
 
-            while (time() - start_time) < timeout_duration:
+            while consume_preamble_events or (time() - start_time) < timeout_duration:
                 # Handle default events like timeout, host_test_name, ...
                 try:
                     (key, value, timestamp) = event_queue.get(timeout=1)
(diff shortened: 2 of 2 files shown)


jupe:
Please add more description why this changes - what was the problem and how this tries to fix it.

alekla01:
**master:**
  * **problem:** run_test sends __host_test_finished event to conn_process after timeout_duration (~polling_timeout) if preamble (setup phase of conn_process) has not finished in timeout_duration. run_test joins conn_process. conn_process receives __host_test_finished after preamble (allocate, flash, sync, etc) and exits (even if nothing is wrong).
  * **other:** Each preamble component should have their own timeouts if required (such as polling timeout, or allocation timeout). if something goes wrong in conn_process during preamble, conn_process exits with __notify_conn_lost event, run_test receives __notify_conn_lost event and exits the while loop. run_test cannot (at least currently) realistically terminate conn_process (without somewhat significant problems, related to such as allocation and flashing).

```python
-> execute
|| parallel

0 sec: run_test  -> start process conn_process()
0 sec: run_test  -> while timeout_duration: handle events || conn_process -> preamble(allocate, flash, sync, etc..)
60 sec: run_test -> timeout, send __host_test_finished    || conn_process -> preamble
60 sec: run_test -> join(conn_process)                    || conn_process -> preamble
80 sec: run_test -> joined                                || conn_process -> preamble
90 sec: run_test -> joined                                || conn_process -> preamble done, ready to begin testing.
90 sec: run_test -> joined                                || conn_process -> oh, received __host_test_finished, well lets not run the tests then.
90 sec: run_test -> result=timeout                        || conn_process -> exit
```

---

**solution:**
  * **what:** Remove/disable timeout during preamble.
  * **how:** in the (run_test) while loop only check timeout_duration if consume_preamble_events=False, i.e., executing tests, otherwise (consume_preamble_events=True) wait for further events (such as __host_test_name, which sets consume_preamble_events=False) from conn_process.

```python
-> execute
|| parallel

0 sec: run_test   -> start process conn_process()
0 sec: run_test   -> handle events                         || conn_process -> preamble(allocate, flash, sync, etc..)
60 sec: run_test  -> handle events                         || conn_process -> preamble
80 sec: run_test  -> handle events                         || conn_process -> preamble
90 sec: run_test  -> handle events                         || conn_process -> preamble done, ready to begin testing.
90 sec: run_test  -> timeout_duration=test_case_timeout    || conn_process -> begin_testing
90 sec: run_test  -> while timeout_duration: handle events || conn_process -> testing, send events
100 sec: run_test -> result=test_result                    || conn_process -> tests executed, exit
```

The solution is not ideal, but should work better than the current implementation.

theotherjimmy:
@bridadan Does this seem like an okay fix to you. I'm not sure I'm wiling to drop a timeout...

bridadan:
I don't think removing the timeout is the right fix. If the process hangs indefinitely it could seriously hurt CI processes.

From the JIRA issue you mentioned, I suggested just increasing the polling timeout (`-P` or `--polling-timeout`). Would that not be enough in this situation?

Issue #206, opened 2018-07-16
Traceback on invalid character


c1728p9:
```
[1531774497.10][CONN][RXD] >>> 'Test attach_us for 500ms and time measure': 1 passed, 0 failed
[1531774497.10][CONN][RXD]
[1531774497.13][CONN][RXD] >>> Running case #9: 'Test detach'...
Process Process-1:
Traceback (most recent call last):
  File "c:\python27\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "c:\python27\lib\multiprocessing\process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "c:\python27\lib\site-packages\mbed_host_tests\host_tests_conn_proxy\conn_proxy.py", line 257, in conn_process
    print_lines = kv_buffer.append(data)
  File "c:\python27\lib\site-packages\mbed_host_tests\host_tests_conn_proxy\conn_proxy.py", line 45, in append
    self.buff += payload.decode('utf-8')
  File "c:\python27\lib\encodings\utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xff in position 0: invalid start byte
```

ciarmcom:
ARM Internal Ref: MBOTRIAGE-1280

PR #207, opened 2018-07-19
add try/exception to handle UnicodeDecodeError

aashishc1988:
Due to lack of pull-up on different parts, the re-initialization of uart causes it to send an invalid character which we are not able to handle properly on htrun side. While partners are investigating the issue and fix, we will add this change in to handle the exception gracefully, as this bug is blocking the addition of multiple parts to our CI

### DIFF
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -42,7 +42,10 @@ class KiViBufferWalker():
 
     def append(self, payload):
         """! Append stream buffer with payload and process. Returns non-KV strings"""
-        self.buff += payload.decode('utf-8')
+        try:
+            self.buff += payload.decode('utf-8')
+        except UnicodeDecodeError:
+            pass
         lines = self.buff.split('\n')
         self.buff = lines[-1]   # remaining
         lines.pop(-1)
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -47,6 +47,7 @@ class KiViBufferWalker():
             self.buff += payload.decode('utf-8')
         except UnicodeDecodeError:
             logger.prn_wrn("UnicodeDecodeError encountered!")
+            self.buff += payload.decode('utf-8','ignore')
         lines = self.buff.split('\n')
         self.buff = lines[-1]   # remaining
         lines.pop(-1)
--- a/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
+++ b/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
@@ -46,7 +46,7 @@ class KiViBufferWalker():
         try:
             self.buff += payload.decode('utf-8')
         except UnicodeDecodeError:
-            logger.prn_err("UnicodeDecodeError encountered!")
+            logger.prn_wrn("UnicodeDecodeError encountered!")
         lines = self.buff.split('\n')
         self.buff = lines[-1]   # remaining
         lines.pop(-1)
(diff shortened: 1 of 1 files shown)


aashishc1988:
@cmonr Added a print statement in case unicodeDecode error occurs.

mbed_host_tests/host_tests_conn_proxy/conn_proxy.py
```diff
@@ -42,7 +42,11 @@ def __init__(self):
 
     def append(self, payload):
         """! Append stream buffer with payload and process. Returns non-KV strings"""
-        self.buff += payload.decode('utf-8')
+        logger = HtrunLogger('CONN')
+        try:
+            self.buff += payload.decode('utf-8')
+        except UnicodeDecodeError:
+            logger.prn_err("UnicodeDecodeError encountered!")
```
bridadan: Total nit pick here: I'd go with a `prn_wrn` here since the `htrun` will continue to run.

aashishc1988: logger.prn_err seemed like an obvious choice at first, but then looking at other places where it is used, we do return/raise an exception when we log the message with prn_err, but here we won't be doing anything like that, so I see your point. I will change that.

cmonr:
@aashishc1988 Something I was wondering.

Since the issue is known to be target specific, and a patch for the target appears to be feasable, why not open an Mbed OS PR instead? I'd like to echo @theotherjimmy sentiment about temporary, target-specific tool patches (https://github.com/ARMmbed/greentea/pull/278#issuecomment-407175402) in that I think we'd be much better off if we can avoid them.

bridadan:
For what its worth, I think htrun should be able to handle these unicode errors, even if a patch lands in the target. If you want a patch to land in Mbed OS too then no worries, but I don't mind htrun being cautious here as well.

studavekar:
I agree with @bridadan test runner should handle Unicode character and should not crash. It effectively fixes htrun issue https://github.com/ARMmbed/htrun/issues/206.  

There is an issue on mbed-os https://github.com/ARMmbed/mbed-os/issues/7511 for ST device, we can add other targets which show this behaviour.

theotherjimmy:
how about decode with ignore or decode with replacement? that way we only drop/replace the offending char.

aashishc1988:
@theotherjimmy using decode with ignore/replace will mask the error, I am not entirely sure what we plan on doing with this sort of behavior in future but it would be good to have some sort of message printed out if/when we encounter this issue. What if tomorrow we add a new part which has the same issue, but now since we are masking it, we wouldnt be able to know and push the partners for the fix.

Also, regarding the solution to treat it as bytes object, wouldnt that require change on both greentea and htrun? If you thinkthat is the most apt solution, lets take this offline and discuss what all need to change and the effects of it. I am still fairly new to GT and htrun architecture :)

theotherjimmy:
@aashishc1988 Could you have the except do a decode with ignore? that way we minimize the lost bytes.

aashishc1988:
Yes, thats a good point. We can print a UnicodeDecodeError exception message, and "decode with ignore" as well. Made the change.

theotherjimmy:
@cmor care to review?

cmonr:
I like it! Catches the error, indicates that one happened, and still tries to return the data.

Commit message could be shorter/multiline, but I'm not going to block on a nit.

jeromecoutant:
Hi
Is it possible to release this patch in a version ?
Thx

cmonr:
@jeromecoutant We'll likely be making another release sometime mid next month (if not sooner) with the release of Mbed OS 5.10. There are one or two fixes looking to come in that will need multiple tools to be updated in lock step.

dhwalters423:
@MarceloSalazar this is blocking support for the UBLOX ODIN W2 and SMCC tests.

aashishc1988:
@dhwalters423 Just fyi, this is not target specific. We have seen it on platforms other than ST (I cannot recall names, would need help of a colleague who is out today). 
Also, it looks like the current PR is merged. Are you still seeing an issue?

dhwalters423:
When installing newest htrun from sources I do not see the issue anymore. thanks for clarification.

PR #216, opened 2018-08-29
Update htrun fastmodel connector to match with newer fm_agent 

jamesbeyond:
**add delicate `reset()` function for fast models**
the `reset()` and `run()` were considered to be the same, but in some cases they are different

**change the create to setup function to match fm_agent 1.1**
fm_agent been updated to version 1.1, slight chnages about how to `create()` or `setup()` a model. This PR just to match with the new changes. Also fm_agent module are backwards compatible


jamesbeyond:
Hey guys, any progress on this? @bridadan @theotherjimmy 
I really hope this could go live along with 5.10 release

PR #218, opened 2018-09-28
Rename most modules to reduce redundancy in import paths

### LINKED ISSUE #173: All of the sub-modules start with `host_tests`
This is very very redundant. I'm not sure if anyone even imports them.

theotherjimmy:
Resolves #173


studavekar:
@theotherjimmy thanks for bringing these changes. quick try of htrun as an independent entity works well(locally and with --grm), however integration with greentea looks  broken 

```
[mbed] Exec "mbedgt --test-spec ./BUILD/tests/K64F/GCC_ARM/test_spec.json -n tests-mbed_drivers-lp_ticker -V" in "/home/arm/repos/mbed-os"
Traceback (most recent call last):
  File "/home/arm/repos/envhtrun/bin/mbedgt", line 11, in <module>
    load_entry_point('mbed-greentea', 'console_scripts', 'mbedgt')()
  File "/home/arm/repos/greentea/mbed_greentea/mbed_greentea_cli.py", line 212, in main
    copy_methods_str = "Plugin support: " + ', '.join(mbed_host_tests.host_tests_plugins.get_plugin_caps('CopyMethod'))
AttributeError: 'module' object has no attribute 'host_tests_plugins'
[mbed] ERROR: "mbedgt" returned error.

```

theotherjimmy:
Yeah, greentea needs to be updated in tandem. The API containing "host_test[s]" after literally every "." is annoying.

theotherjimmy:
Also, this got the thinking that The `copy_` and `reset_` plugins are not loaded by filesystem, so they could be in the same file. That could make the code base 1 step easier to navigate, as all "jlink" stuff is in one file.

theotherjimmy:
OTOH, I could make aliases at the root for the old imports...

theotherjimmy:
I'm going to make aliases.

theotherjimmy:
@bridadan Aliases added.

theotherjimmy:
I did not move "host_tests_toolbox" because I still don't know what that's supposed to be.

theotherjimmy:
It seems neither "host_tests" or "toolbox" related.

theotherjimmy:
Ugh. This is why you don't put code into `__init__.py`

theotherjimmy:
@bridadan What does the `host_tests_toolbox` module do in 1-2 words?

bridadan:
> What does the host_tests_toolbox module do in 1-2 words?

😮 
"Flash/Reset" - 2 words!

I don't actually know, I can't find anything that uses it (except Mbed CLI). Looks like it was added as part of this big refactor PR: https://github.com/ARMmbed/htrun/pull/65 But also looks like it was never used.

theotherjimmy:
Actually, `mbed_host_test.flash_reset` is a better name: at least I know what's going on.

Would it make more sense.

bridadan:
Yeah probably, though I don't think we should rename it quite yet, just because we don't know if other things are depending on it. We could revisit it in another PR?

theotherjimmy:
We certainly can revisit it.

bridadan:
Tracked here: https://github.com/ARMmbed/htrun/issues/220

theotherjimmy:
Thanks for opening the tracking issue.

theotherjimmy:
Leaving this for another day.

PR #228, opened 2018-11-20
Feature: dynamic timeouts

screamerbg:
This minimalistic PR adds support to dynamically sett/adjust a test timeout via DUT messages `{{__timeout_set,int}}`, `{{__timeout_adjust,int}}`.

The is particularly useful when a device is making a download (say firmware update), where the speed might dynamically change based on various conditions. With this feature, the DUT might notify the host that it's alive and running, just slower than expected.

A typical usecase would be a unified IP connectivity test, where the time to complete over Ethernet and Wifi would be significantly faster than 2G/3G, let alone NBIoT/CAT-M1.

As discussed with @bridadan

### DIFF
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -437,7 +437,7 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                         # Dynamic timeout set
                         timeout_duration = int(value) # New timeout
                         self.logger.prn_inf("setting timeout to: %d sec"% int(value))
-                    elif key == '__timeout_adj':
+                    elif key == '__timeout_adjust':
                         # Dynamic timeout adjust
                         timeout_duration = timeout_duration + int(value) # adjust time
                         self.logger.prn_inf("adjusting timeout with %d sec (now %d)" % (int(value), timeout_duration))
--- a/mbed_host_tests/host_tests_runner/host_test_default.py
+++ b/mbed_host_tests/host_tests_runner/host_test_default.py
@@ -433,6 +433,14 @@ class DefaultTestSelector(DefaultTestSelectorBase):
                         self.logger.prn_inf("%s received"% (key))
                         callbacks__exit_event_queue = True
                         break
+                    elif key == '__timeout_set':
+                        # Dynamic timeout set
+                        timeout_duration = int(value) # New timeout
+                        self.logger.prn_inf("setting timeout to: %d sec"% int(value))
+                    elif key == '__timeout_adj':
+                        # Dynamic timeout adjust
+                        timeout_duration = timeout_duration + int(value) # adjust time
+                        self.logger.prn_inf("adjusting timeout with %d sec (now %d)" % (int(value), timeout_duration))
                     elif key in callbacks:
                         # Handle callback
                         callbacks[key](key, value, timestamp)


theotherjimmy:
@screamerbg Thanks for the PR. Do you need this in a release of htrun soon?

screamerbg:
@theotherjimmy Yes please (if possible).

Here's an example usage - https://github.com/ARMmbed/simple-mbed-cloud-client/pull/49/commits/6cb6566c550c590498b8e32b1d504629006e4b90#diff-82f388e30987a4563c237f17ccb185dbR66

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -433,6 +433,14 @@ def process_code_coverage(key, value, timestamp):
                         self.logger.prn_inf("%s received"% (key))
                         callbacks__exit_event_queue = True
                         break
+                    elif key == '__timeout_set':
```
bridadan: The beginning of the test (the "preamble" as its referred to at the top of this file) sets the timeout of the test with the `__timeout` key. I'd say go ahead and reuse that key instead of making a new one, unless you have any objections.

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -433,6 +433,14 @@ def process_code_coverage(key, value, timestamp):
...
+                        # Dynamic timeout set
+                        timeout_duration = int(value) # New timeout
+                        self.logger.prn_inf("setting timeout to: %d sec"% int(value))
+                    elif key == '__timeout_adj':
```
bridadan: I tend to stay away from abbreviations if the full word is reasonably small. In this case `__timeout_adjust` is just three more characters. Would you mind "adjusting" (heh, couldn't resist 😄) this key?

mbed_host_tests/host_tests_runner/host_test_default.py
screamerbg: The `__timeout` key at "preamble" stage also sets the global timer to 0 and timeout duration. The behavior of `__timer_set` only sets the timeout duration, but doesn't change the global timer (e.g. time elapsed). I'd prefer not to confuse these two different behaviors.

mbed_host_tests/host_tests_runner/host_test_default.py
screamerbg: Sure.

mbed_host_tests/host_tests_runner/host_test_default.py
bridadan: That's a very good point, thanks very much for clarifying 😄

screamerbg:
Bump @cmonr @theotherjimmy

mbed_host_tests/host_tests_runner/host_test_default.py
```diff
@@ -433,6 +433,14 @@ def process_code_coverage(key, value, timestamp):
                         self.logger.prn_inf("%s received"% (key))
                         callbacks__exit_event_queue = True
                         break
+                    elif key == '__timeout_set':
+                        # Dynamic timeout set
+                        timeout_duration = int(value) # New timeout
+                        self.logger.prn_inf("setting timeout to: %d sec"% int(value))
+                    elif key == '__timeout_adjust':
+                        # Dynamic timeout adjust
+                        timeout_duration = timeout_duration + int(value) # adjust time
```
cmonr: I was about to ask "what's the difference between `adjust` and `set`, but this was simple enough :)

